PR #17, opened 2019-11-07
Shared colors

davidpauljunior:
Imports the shared colours from global.
Changes `colors` to `default`.


davidpauljunior:
Yeah that was so satisfying! 👏

PR #25, opened 2019-11-13
Bumps peer dependency springer-context to 12.2.2 on outdated packages

ricky-lau:
- updated in springer-author-list, springer-box, springer-lists
- updated major version of each package
- some formatting change to be consistent

### DIFF
--- a/toolkits/springer/packages/springer-author-list/package.json
+++ b/toolkits/springer/packages/springer-author-list/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@springernature/springer-author-list",
-  "version": "0.1.0",
+  "version": "1.0.0",
   "license": "MIT",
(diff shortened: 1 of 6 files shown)


toolkits/springer/packages/springer-author-list/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@springernature/springer-author-list",
-  "version": "0.1.0",
+  "version": "1.0.0",
```
davidpauljunior: Is this a breaking change for author-list?

ricky-lau: Going to bump it to 0.2.0 instead and for the others to minor rather than major version as there were no breaking changes 😬

PR #28, opened 2019-11-19
Updates filenames for consistency across codebase

benjclark:
we had a mix of `_filename.scss` and `filename.scss`

we also had a mix of `_toolkitName-componentName.scss` and `_componentName.scss`

this pr standardises stuff across the toolkits

spoke to alex kilgour before doing this. haven't changed context filenames (e.g. `/springer-context/`) to be underscored as there is a special reason for those not being underscored. but yeah, besides that it is best practice to underscore sass files that do not end up as standalone stylesheets that are used in the browser

might add these as linting rules to the frontend-package-manager in the future


davidpauljunior:
I think this should be from `_mything` to `mything`

toolkits/springer/packages/springer-context/scss/enhanced.scss
```diff
@@ -41,13 +41,13 @@
 @import '30-mixins/headings';
 @import '30-mixins/triangle';
 @import '30-mixins/wrap';
-@import '30-mixins/springer-hiding';
-@import '30-mixins/springer-links';
-@import '30-mixins/springer-typography';
+@import '30-mixins/hiding';
```
ricky-lau: I think this bit suppose to keep the `springer-` prefix? you cant tell the difference between importing `links` and the other two from global and springer toolkits

davidpauljunior:
I think we may as well keep it as no underscores everywhere for consistency.
The reason for using underscores in scss is because when you have a glob like `**/*.scss`, it will only compile the files that are not underscored.
But we can't use that approach anyway if some of the files aren't underscored.

toolkits/springer/packages/springer-context/scss/enhanced.scss
benjclark: ah yeah i always forget there's magic happening here isn't there. how does this work again? if there's a global one with matching filename it will load that in first and will then overwrite it with springer one?

davidpauljunior: If there are any matching filenames between the brand context and the global context, the brand file will win.  The contents won't be magically merged, it just takes the brand one as gospel.

benjclark: ah ok, so we want to keep the filenames the same in brand and global then so that the logic works. therefore we should not prefix the filenames with springer. because that would undermine the underlying filename matching logic entirely, making it redundant. right?

ricky-lau: hrmm, I guess that means the contents of `links` from global should be copied into the brand level (I added that one so that's why I keep going on about it 😁)

davidpauljunior: We prefix the name with Springer when we need some code in addition to what's in global.
It's like saying, we need the global hiding stuff, but we also have our own hiding stuff that isn't global.

davidpauljunior: Looking at what's in the files, `u-visually-hidden-focus` can go into global (no reason no to IMO).
Once that's gone, the springer-hidden makes sense, it just adds `class-stack` stuff which probably shouldn't be in global.
So we're saying global has a file called `hiding` and then we have some brand specific stuff which is in a file called `springer-hiding`

benjclark: ah now i get you, the brand prefixed ones extend the global and we don't want to duplicate code across both toolkits so extending is important. i guess it's going to be pretty rare that we need the magic logic that chooses a brand one over a global one, we'd only need to do that if we want to break away from the global component entirely for some special reason. strange.

thanks for your explanations. i'm going to close this and get on with story work and probably start a new pr to tackle bits of this again from scratch.

PR #67, opened 2020-01-17
global-banner: change to layout and documentation

davidpauljunior:
- Removes pseudo elements that were previously used to extend the banner to the edge of the screen
- Updates the documentation with information on using app level containers to control the content
- Adds horizontal padding and a flush modifier to allow overriding of it


toolkits/global/packages/global-banner/package.json
```diff
@@ -1,8 +1,8 @@
 {
   "name": "@springernature/global-banner",
-  "version": "0.1.1",
+  "version": "0.2.0",
   "license": "MIT",
-  "description": "Represents a box with text",
+  "description": "Full width banner used for displaying informational messages",
   "keywords": [],
   "author": "Springer Nature",
   "peerDependencies": {
```
ricky-lau: not related to our changes but could bump this to 10.0.0 😁

davidpauljunior:
That's actually a mistake. I was going to leave `marketing` but change `inverted` to `contrast`

davidpauljunior:
Then I was hoping to change across all projects. And then release this as a major version.

toolkits/global/packages/global-banner/package.json
davidpauljunior: ^^ global-context 😄

ricky-lau: 🤦‍♂that's what i meant

PR #69, opened 2020-01-17
springer-webfont: Remove throws related to sessionStorage

davidpauljunior:


toolkits/springer/packages/springer-webfonts/js/index.js
```diff
@@ -1,5 +1,11 @@
 import FontFaceObserver from 'fontfaceobserver';
 
+/**
+ * 1. No need to throw this.
+ *	  Access to browser storage can be blocked in the browser settings and that is a valid user choice.
+ *	  The try catch is required because sessionStorage will still be defined and JavaScript will error.
```
benjclark:
```suggestion
 *	  The try catch is required because sessionStorage will still be defined and JavaScript will error but we want it to fail silently.
```

benjclark:
Hooray for comments! 🎉  I'm all for more commenting

PR #83, opened 2020-01-23
allow brand settings separate to global

nickcall:


davidpauljunior:
Are the deleted vars from here the ones that are being set in global?

nickcall:
Yeah thats right, they were the same in global as in brand so I removed them from brand, is that correct ?

davidpauljunior:
Yep!

PR #84, opened 2020-01-27
Update extends version

nickcall:


toolkits/nature/packages/nature-article/HISTORY.md
```diff
@@ -1,5 +1,8 @@
 # History
 
+## 0.4.1 (2020-01-27)
+	* Update the extends package version of glbal article from 7.2.0 to 9.3.0
```
saramorueta: missing 'o'

toolkits/nature/packages/nature-article/package.json
```diff
@@ -1,10 +1,10 @@
 {
   "name": "@springernature/nature-article",
-  "version": "0.4.0",
+  "version": "0.4.1",
```
saramorueta: I'd say this is a major version because we might be including a lot of new stuff right?

nickcall: yeah I guess, will up date. thanks.

davidpauljunior: You don't really need to major version if it's still on version 0 (development).
If this is used in production then it should _already_ be a major version.

nickcall: Have pushed it up to 0.5.0, is that OK ?

nickcall: I should really have done this as part of the previous PR

davidpauljunior:
In this PR it doesn't really matter in my opinion.
However, if this article component is used in production and changes could result in breakages then a major version should be released (same for the global article).
That can be done in separate PRs.

PR #98, opened 2020-02-10
springer-webfonts: first major release

davidpauljunior:


toolkits/springer/packages/springer-webfonts/HISTORY.md
```diff
@@ -1,3 +1,6 @@
+## 1.0.0 (2019-02-10)
```
nickcall: 2020 ??

davidpauljunior: Oh. I wonder how many packages I've done that for 🤔

davidpauljunior: At least the one before on this package it would seem

nickcall: I probably signed it off

PR #99, opened 2020-02-10
Breaking: Remove webfont usage from springer-context

davidpauljunior:
Springer branding doesn't use webfonts so has no need for fallbacks and webfont mixins.


toolkits/springer/packages/springer-context/scss/10-settings/typography.scss
```diff
@@ -0,0 +1,35 @@
+$font-family-sans: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen-Sans, Ubuntu, Cantarell, 'Helvetica Neue', sans-serif;
```
benjclark: Are these sans and serif variables going to be become the default for all products? Would it make sense to set these in the global context and override for Nature or wherever that doesn't use

nickcall: They are already in the global-context, https://github.com/springernature/frontend-toolkits/blob/master/toolkits/global/packages/global-context/scss/10-settings/typography.scss

davidpauljunior: Yeah, the sans is actually already in the global.
After this PR, Nick is going to work on only setting the overrides per brand.

benjclark: If they are already in global then I guess they aren't needed here in springer context then? as you are not overriding the values

davidpauljunior: It doesn't merge the contents of the files. I could rename this to `typography-overrides.scss`

nickcall: yeah i guess that the better naming convention, I could update nature to be the same.

davidpauljunior: Won't be needed when we roll out the new system.
I'll change to `typography-overrides.scss`.  Doesn't _always_ override but it's probably close enough for a temporary solution.

davidpauljunior: Or we could go with `typography-springer.scss`

davidpauljunior: Ok that's done. Will need a re-review.

PR #113, opened 2020-02-12
global-banner: sets underline on links

davidpauljunior:
Banner links should always have underlines, regardless of the brand.
This is because the banner is intended to be a combination of a text and links (normally in a sentence).

Bumps global-context peer dep.


davidpauljunior:
@nickcall I might as well bring this up to version 13 of the context.

PR #132, opened 2020-02-17
Add new style for article body

nickcall:
Trying to use the % width div container across all shared article pages (BMC,SL,SO,Nature)


toolkits/global/packages/global-article/scss/50-components/article-enhanced.scss
```diff
@@ -3,6 +3,16 @@
  * Article styles for browsers that pass CTM
  */
 
+.c-article-main-column {
```
davidpauljunior: This would be _slightly_ less code if we did mobile first.

```
.c-article-main-column {
    width: 100%; // This may even be redundant

	@include media-query(map-get($article--breakpoints, 'med')) {
		width: 60.2%;
		margin-right: 8.6%;
	}
}
```

PR #154, opened 2020-02-24
Global autocomplete: Enhance dynamic data support

morgaan:
- Enhance arguments with additional `httpMethod` setting
- Enhance arguments with additional `bodyTemplate` setting
- Add support for the two above new settings
- Delegate "No results" to resultsCallBack rather than generateSuggestions
- Update demo
- Update HISTORY
- Bump to next major version because of breaking change

### DIFF
--- a/toolkits/global/packages/global-autocomplete/__tests__/index.js
+++ b/toolkits/global/packages/global-autocomplete/__tests__/index.js
@@ -133,16 +133,16 @@ describe('Autocomplete', () => {
 		});
 
 		describe('A response with zero suggestions', () => {
-			test('Should add "no results" to the list', async () => {
+			test('Should not error out because the "no results" situation is the job of the resultsCallBack', async () => {
 				let auto = autoComplete(args);
 				auto.enable();
 
-				setMockFetch({ok: true, status: 200, json: () => { return [] }});
+				setMockFetch({ok: true, status: 200, json: () => { return []; }});
 				input.value = 'bur';
 				input.dispatchEvent(new Event('keyup'));
 
 				await waitFor(2);
-				expect(mockResultsCallback).toHaveBeenCalledWith(['No results']);
+				expect(mockResultsCallback).toHaveBeenCalledWith([]);
 			});
 		});
 });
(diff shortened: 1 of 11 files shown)


toolkits/global/packages/global-autocomplete/__tests__/index.js
```diff
@@ -56,20 +56,42 @@ describe('Autocomplete', () => {
...
 				input.value = 'burdmen';
 				input.dispatchEvent(new Event('keyup'));
 
 				await waitFor(2);
-				expect(fetchSpy).toHaveBeenCalledWith(url + 'burdmen', {"content-type": "application/json", "headers": headers});
+				expect(fetchSpy).toHaveBeenCalledWith(url + 'burdmen', {"content-type": "application/json", "headers": headers, "method": 'GET'});
 
 			});
 
+			test('Or be able to send a request to the configured endpoint, HTTP method and bodyTemplate', async () => {
```
jon-stevens:
```suggestion
			test('Should send a request using the configured HTTP method and bodyTemplate', async () => {
```

toolkits/global/packages/global-autocomplete/js/index.js
```diff
@@ -120,8 +123,11 @@ const autoComplete = arguments_ => {
 
 	const generateSuggestions = data => {
 		removeSuggestions();
+		data = toSuggestionsCallback(data);
```
jon-stevens: I would suggest this be optional, as the server might already be returning an array, or the consumer might be using static data

jon-stevens: I would use a new reference if you are modifying the data and maybe call it something like suggestions?

```suggestion
		const suggestions = toSuggestionsCallback(data);
```

toolkits/global/packages/global-autocomplete/js/index.js
```diff
@@ -11,8 +11,11 @@ const autoComplete = arguments_ => {
 		inputDelay = 300,
 		requestTimeout = 2000,
 		headers = {},
+		httpMethod = 'GET',
+		bodyTemplate,
```
jon-stevens: This is cool, could you maybe add it to the demo stuff so that it is documented, otherwise it might not be clear

morgaan: I will, very good point. I though about the tests but not the demo 🤦‍♂️

morgaan:
It has been forked to #160

PR #157, opened 2020-02-24
nature-context: nature headings

davidpauljunior:
Heading styles applied via mixins.
`h4` is the final level. `h5` and `h6` are presentationally the same as `h4`.

Note: I could find anything that uses the `h5` tag in the html for Nature so it felt safe to remove.


sky-jack:
ah ok utilities live elsewhere, that makes sense.

davidpauljunior:
@sky-jack yep you're right, and the reason the utilities live elsewhere is because it's up to the application to import them only if they want them. That way we don't bloat css unnecessarily.

PR #162, opened 2020-02-26
removes element level styles in global forms package 

sky-jack:
and adds styles to textarea component modifier


toolkits/global/packages/global-forms/scss/30-mixins/_global-form-input.scss
```diff
@@ -12,11 +12,13 @@
 	$border-radius: $global-form--border-radius
 ) {
 	width: 100%;
+	outline: 0;
```
davidpauljunior: Are we setting any `:focus` styles on these inputs?

alexkilgour: I do think that needs addressing, but I think for the purpose of this PR, Jack is trying to get rid of those base styles as per the discussion in #frontend, I will make an issue for the focus stuff

sky-jack: Focus styles for forms are currently living at the brand level. Perhaps we do need some global defaults or unity across brands so yes definitely something to think about

PR #173, opened 2020-03-02
convert InternalNavigation to use module.exports and add test coverage

moddular:


toolkits/global/packages/global-article/js/internal-navigation.js
```diff
@@ -1,151 +1,151 @@
 
-window.Component.InternalNavigation = (function (win, document_) {
+var InternalNavigation = (function (window, document) {
```
saramorueta: is this IIFE really necessary? I'm asking because if we keep it I think we would need to instantiate this component differently as we expect to have just a named function instead. 
 
https://github.com/springernature/oscar/blob/bc718afc64ceee4c8ba20ae0222ada43a24864d6/oscar-sites-nature/oscar-mosaic/js/nature.js#L41

PR #179, opened 2020-03-03
Catches font face observer rejected promises correctly

benjclark:
We've been getting a shed load of timeout errors sent through to airbrake. It happens when a user with a slow connection executes the font face observer js we have on BMC as it has a 3sec timeout before erroring.

This catches the rejected promise in the correct place and swallows the error 

image


davidpauljunior:
Think you may need to mock `observer`?

PR #187, opened 2020-03-10
Remove dependency on jQuery from Scheduler, make available through mo…

moddular:
…dule.exports

### DIFF
--- a/toolkits/global/packages/global-article/__tests__/scheduler.js
+++ b/toolkits/global/packages/global-article/__tests__/scheduler.js
@@ -1,0 +1,226 @@
+const Scheduler = require('../js/scheduler');
+
+let readyState = 'complete';
+
+Object.defineProperty(document, 'readyState', {
+	get() { return readyState; }
+});
+
+describe('Scheduler', () => {
+	let scheduler;
+	let addEventListenerSpy;
+	let removeEventListenerSpy;
+	let requestAnimationFrameSpy;
+	let cancelAnimationFrameSpy;
+
+	beforeEach(() => {
+		addEventListenerSpy = jest.spyOn(window, 'addEventListener');
+		removeEventListenerSpy = jest.spyOn(window, 'removeEventListener');
+		requestAnimationFrameSpy = jest.spyOn(window, 'requestAnimationFrame');
+		cancelAnimationFrameSpy = jest.spyOn(window, 'cancelAnimationFrame');
+		scheduler = Scheduler();
+	});
+
+	afterEach(() => {
+		addEventListenerSpy.mockRestore();
+		removeEventListenerSpy.mockRestore();
+		requestAnimationFrameSpy.mockRestore();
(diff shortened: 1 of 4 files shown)


toolkits/global/packages/global-article/__tests__/scheduler.js
```diff
@@ -0,0 +1,226 @@
+const Scheduler = require('../js/scheduler');
```
benjclark: Is there any reason why you're using `require` and `module.exports` instead of `import` and `export`. I think the latter has become the norm in other toolkit components?

benjclark: I meant that for both components and for their tests

benjclark: Actually, I've just double checked myself on that, and it seems we have a mix of both styles in the toolkits. Hoorah! 😂

moddular: Yeah, I think global-article has only been using `module.exports` to this point. I don't know if we switch whether to do them in bulk or one-by-one?

benjclark: Yeah I guess converting one or a few at a time is more sensible and could be done as a standalone task. Probs not worth worrying about as part of this PR. The playbook suggests `import` and named `exports` is the way we should be doing it. https://github.com/springernature/frontend-playbook/blob/master/javascript/house-style.md

saramorueta: well.. the issue with this is that the consumer part of these components on nature cannot import ES6 modules yet. That will come later, as soon as we start migrating code.

benjclark: Oh does the consumer not process the js bundle using babel then?

saramorueta: Only the new bundle `nature-es6.js` can do. The idea is to migrate _gradually_ all the code from `nature.js` to the new `nature-es6.js` to start using ES6 modules. 
So far, the only component that has been included in that new bundle is `share-box`. So that's the only one that potentially can be exported to ES6.

benjclark: Oh it's because the consumer of this is old mosaic, isn't it? And that uses a ruby syntax for importing js files?

saramorueta: yep! we're still using mincer to create the bundle for nature.

benjclark: Sorry, all of this stuff had fallen out of my brain long ago. Thanks for the explanations

saramorueta: it's a bit confusing we know... :( thanks for asking!

PR #190, opened 2020-03-11
Update springer and nature context 

ricky-lau:
bump extendsPackage global-context to 14.2.1


toolkits/nature/packages/nature-context/scss/10-settings/colors/default.scss
```diff
@@ -24,7 +24,7 @@ $colors: (
 	'steel': #bcd2dc,
 	'grape': #bee188,
 	'lemon': #fece3e,
-	'orange': #cc4b14,
+	'orange': #b74616,
```
davidpauljunior: Should this call the global setting for open-access?
`orange: $open-access-color`
Should it be here at all if open access color is set at the global level (https://github.com/springernature/frontend-toolkits/blob/master/toolkits/global/packages/global-context/scss/10-settings/colors/shared.scss)

PR #191, opened 2020-03-11
Dropdown component accept array of classnames as option

benjclark:
Rather than limiting the dropdown component to only allow one classname to be added when passing in options, you can now pass in an array of classnames should you need.

This is needed in Springer where I will be making multiple dropdowns and they require 2 classnames each.

### DIFF
--- a/toolkits/springer/packages/springer-dropdown/__tests__/unit/dropdown.spec.js
+++ b/toolkits/springer/packages/springer-dropdown/__tests__/unit/dropdown.spec.js
@@ -15,7 +15,7 @@ const selector = {
 const defaultOptions = {
 	BUTTON_CLASS: '',
 	CLICK_OUTSIDE: true,
-	DROPDOWN_CLASS: '',
+	DROPDOWN_CLASSES: [''],
 	HIDE_CLASS: 'u-hide',
 	HIDE_INITIALLY: true,
 	MENU_POSITION: 'left'
@@ -307,9 +307,9 @@ describe('springer-dropdown', () => {
 
 	describe('configuration options', () => {
 		describe('via function parameter', () => {
-			test('DROPDOWN_CLASS: Should add additional className to dropdown', () => {
+			test('DROPDOWN_CLASSES: Should add additional className to dropdown', () => {
 				// When
-				dropdown({DROPDOWN_CLASS: 'additional-dropdown-class'});
+				dropdown({DROPDOWN_CLASSES: ['additional-dropdown-class']});
 
 				// Then
 				expect(element.DROPDOWN.classList.contains('additional-dropdown-class')).toBe(true);
@@ -356,7 +356,7 @@ describe('springer-dropdown', () => {
 		});
 
 		describe('via data-* attributes', () => {
-			test('DROPDOWN_CLASS: Should add additional className to dropdown', () => {
+			test('DROPDOWN_CLASSES: Should add additional className to dropdown', () => {
 				// Given
 				element.BUTTON.setAttribute('data-dropdown-dropdown-class', 'additional-dropdown-class');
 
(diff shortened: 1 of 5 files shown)


toolkits/springer/packages/springer-dropdown/js/dropdown.js
```diff
@@ -11,7 +11,7 @@ const selector = {
 const defaultOptions = {
 	BUTTON_CLASS: '',
 	CLICK_OUTSIDE: true,
-	DROPDOWN_CLASS: '',
+	DROPDOWN_CLASSES: [''],
```
davidpauljunior: Should this default to an empty array? This will have a length of 1.

toolkits/springer/packages/springer-dropdown/js/dropdown.js
```diff
@@ -171,8 +171,14 @@ const Dropdown = class {
 			this._menuEl.classList.add(this._options.HIDE_CLASS);
 		}
 
-		if (this._options.DROPDOWN_CLASS.length > 0) {
-			this._dropdownEl.classList.add(this._options.DROPDOWN_CLASS);
+		if (this._options.DROPDOWN_CLASSES.length > 0) {
+			// for each classname passed in
+			for (let i = 0; i < this._options.DROPDOWN_CLASSES.length; i++) {
```
davidpauljunior: Any reason to not use `forEach`?

davidpauljunior: On second thoughts, you don't need a loop.
If the option was `DROPDOWN_CLASSLIST` then it could take comma separated list of classes and you wouldn't need to loop through them. You could just `classList.add(DROPDOWN_CLASSLIST)` (because you can do `classList.add('class1', 'class2', 'class3')`

toolkits/springer/packages/springer-dropdown/__tests__/unit/dropdown.spec.js
```diff
@@ -1,6 +1,7 @@
 import {dropdown} from '../../js';
 import {Dropdown} from '../../js/dropdown';
 import {constants} from '../../js/constants';
+import classPrivateFieldLooseBase from "@babel/runtime/helpers/esm/classPrivateFieldLooseBase";
```
davidpauljunior: What's this for?

toolkits/springer/packages/springer-dropdown/js/dropdown.js
davidpauljunior: Update: see https://github.com/springernature/frontend-toolkits/pull/191/files#r391114098

toolkits/springer/packages/springer-dropdown/__tests__/unit/dropdown.spec.js
benjclark: Where the shit did that come from?!

toolkits/springer/packages/springer-dropdown/js/dropdown.js
benjclark: I tried the comma separated idea. The trouble is how to do you feed them in to classList as arguments, as you can't feed them in as `"class1, class2, class3"`, you need to feed them in as 3 separate strings with comma separating them. I went ahead with that but once I'd written the code to do all that string processing to achieve that it became clear that is simpler and less code to have an array instead that does multiple calls of `classList.add()`.

benjclark: re `forEach()` - that one baffled me. For some reason that code blows up in Jest tests if you use forEach but is fine in the browser. In Jest test output is says forEach does not exist for `this._options.DROPDOWN_CLASSES` even though if you clog the values going in to it it's clearly an array. It was very weird and after an hour I fell back to a for loop.

toolkits/springer/packages/springer-dropdown/js/dropdown.js
benjclark: Ah yes good point

benjclark:
@davidpauljunior As discussed, better implementation in this PR instead: https://github.com/springernature/frontend-toolkits/pull/192

PR #194, opened 2020-03-12
Global menu

davidpauljunior:
Adds a `c-menu` component.

Defaults to horizontal and uses modifiers to make the following variations:

- small (smaller font size)
- inherit (link inherits color)
- vertical (items stack vertically)
- keyline (puts a keyline between items)
- padding (makes links more clickable)


toolkits/global/packages/global-menu/scss/50-components/_menu.scss
```diff
@@ -0,0 +1,112 @@
...
+.c-menu {
+	@include u-list-reset();
+	font-size: $menu--font-size;
+	font-family: $menu--font-family;
+	display: flex;
+	line-height: $menu--line-height;
+}
+
+.c-menu__item:not(:last-child) {
+	margin-right: 16px;
```
benjclark: `get-spacing()` ?

benjclark: or a variable?

davidpauljunior: 👍 good spot. Have updated to use the variables

PR #199, opened 2020-03-17
Remove TabGroup it's no longer used since the metrics pages were rede…

moddular:
…signed


moddular:
I suspect it wasn't designed to be just used on metrics, it just happened that metrics was the only place still using it

PR #201, opened 2020-03-19
Add buttons to global-toolkit

ricky-lau:
Moving button component from springer/bmc/nature150 apps into global


toolkits/global/packages/global-buttons/scss/10-settings/buttons.scss
```diff
@@ -0,0 +1,77 @@
...
+	'hoverColor': white,
+	'hoverBorderColor': greyscale(40),
+	'hoverBackgroundColor': greyscale(40),
+	'hoverBackgroundImage': none,
+	'hoverTextDecoration': none,
+	'focusBorderColor': greyscale(40),
+	'focusTextDecoration': none
+);
+
+$button--ghost: (
```
sky-jack: I'm wondering what the ghost modifier is for?

sky-jack: IE is it a name/distinction that all brands will potentially use?

ricky-lau: It was the name used for buttons that have a transparent background and with visible border on button. I added a line in the readme to describe but can add to that if that was too vague

ricky-lau: Actually some of these themes including ghost probably not be in global and I also need to neutralise the colors so they are not springer specific.

sky-jack: ah ok. yes I wasn't sure if that appearance was going to be global. Maybe we could just specify a primary and secondary style that can be overridden at brand level? Ghost seems a little prescriptive maybe?

ricky-lau: Yep totally agree. I'll have another PR for the brand level stuff after this.

sky-jack: Good stuff 👍

ricky-lau:
I was thinking to use utility classes in the HTML to manage fonts/sizing. But can see it being worthwhile adding font-related styles to the default theme and then have modifiers for a small and large buttons.

I think our buttons across brands are usually sans-serif so maybe leave that for utility classes to override.

sky-jack:
sounds sensible.

toolkits/global/packages/global-buttons/scss/10-settings/buttons.scss
```diff
@@ -0,0 +1,69 @@
+/**
+ *  Button settings
+ *
+ */
+
+$button-base--text-align: left;
```
davidpauljunior: I think the differentiation between base and shared is confusing. You'd have to go into the code to understand which settings to change for your brand.
E.g. shared means it's shared between regular and circle? Which then suggests that other buttons don't use that?

It would read better IMO if it was just `$button--padding`.
Then _if_ we need a circle button, something like `$button--circle-padding: $button--padding` (allows the brand to configure but sets a single default).

toolkits/global/packages/global-buttons/scss/10-settings/buttons.scss
```diff
@@ -0,0 +1,69 @@
...
+$button-base--font-size-normal: 16px;
+$button-base--line-height: 1.4;
+$button-base--font-family: sans-serif;
+$button-base--border-radius: 2px;
+$button-shared--border-width: 1px;
+$button-shared--border-style: solid;
+$button-shared--transition: 0.25s ease, color 0.25s ease, border-color 0.25s ease;
+$button--font-size-small: 14px;
+$button--font-size-large: 18px;
+$button--line-height-small: 1.2;
```
davidpauljunior: Could the line height be shared across the 2 sizes? Like 1.3 or something?

toolkits/global/packages/global-buttons/scss/10-settings/buttons.scss
```diff
@@ -0,0 +1,69 @@
...
+$button-base--text-align: left;
+$button-base--padding: 8px;
+$button-base--font-size-normal: 16px;
+$button-base--line-height: 1.4;
+$button-base--font-family: sans-serif;
+$button-base--border-radius: 2px;
+$button-shared--border-width: 1px;
+$button-shared--border-style: solid;
+$button-shared--transition: 0.25s ease, color 0.25s ease, border-color 0.25s ease;
+$button--font-size-small: 14px;
```
davidpauljunior: rem?

toolkits/global/packages/global-buttons/scss/10-settings/buttons.scss
```diff
@@ -0,0 +1,69 @@
...
+$button--font-size-large: 18px;
```
davidpauljunior: rem?

toolkits/global/packages/global-buttons/scss/10-settings/buttons.scss
```diff
@@ -0,0 +1,69 @@
...
+ *
+ */
+
+$button-base--text-align: left;
+$button-base--padding: 8px;
+$button-base--font-size-normal: 16px;
+$button-base--line-height: 1.4;
+$button-base--font-family: sans-serif;
+$button-base--border-radius: 2px;
+$button-shared--border-width: 1px;
```
davidpauljunior: Can these be combined? `$button--border: 1px solid`?

toolkits/global/packages/global-buttons/scss/10-settings/buttons.scss
davidpauljunior: oh, also I think buttons should default to center text aligned

toolkits/global/packages/global-buttons/scss/10-settings/buttons.scss
davidpauljunior: Personal opinion, I think it reads easier as `$component--modifier-property`.
`$button--small-font-size`

toolkits/global/packages/global-buttons/scss/50-components/buttons.scss
```diff
@@ -0,0 +1,31 @@
...
+.c-button--primary {
+	@include button-theme($button--primary);
+}
+
+.c-button--inline {
+	display: inline-block;
+	width: auto;
+}
+
+.c-button--disabled {
```
davidpauljunior: Should this be `$c-button--disabled, $c-button[disabled]`?

toolkits/global/packages/global-buttons/scss/50-components/buttons.scss
```diff
@@ -0,0 +1,31 @@
...
+	font-size: $button--font-size-large;
+	line-height: $button--line-height-large;
+}
+
+.c-button--primary {
+	@include button-theme($button--primary);
+}
+
+.c-button--inline {
+	display: inline-block;
```
davidpauljunior: I think inline-block makes sense as the default for buttons?
That's how the behave out of the box in the browser.
Then `c-button--block`? 😬

toolkits/global/packages/global-buttons/README.md
```diff
@@ -0,0 +1,40 @@
...
+```html
+<button class="c-button c-button--large">text</button>
+```
+
+**Primary**
+```html
+<button class="c-button c-button--primary">text</button>
+```
+
+**Inline**
```
davidpauljunior: This will need to switch to block 😄

PR #206, opened 2020-03-20
convert reading companion to use module.exports and add tests and export

saramorueta:
The changes on the reading-companion.js correspond mostly to indentation and other bits related to transforming the component in an IIFE function that returns another function. 
And lately, getting the component to export.

There is one test that can be added still. But for now, the tests cover 90% of the code.

### DIFF
--- a/toolkits/global/packages/global-article/__tests__/reading-companion.js
+++ b/toolkits/global/packages/global-article/__tests__/reading-companion.js
@@ -187,7 +187,7 @@ describe('Reading Companion', () => {
 		expect(tabs.children.length).toBe(3);
 	});
 
-	test('Should set sections as default active tab', () => {
+	test('Should set sections as the default active tab', () => {
 		includeMockFigures();
 		includeMockReferences();
 		initReadingCompanion();
@@ -197,7 +197,7 @@ describe('Reading Companion', () => {
 		expect(tabs.querySelector('[data-tab-target="sections"]').getAttribute('aria-selected')).toBe('true');
 	});
 
-	test('Should create list of items which contains a anchor element to the existing abstract section', () => {
+	test('Should create a list of items which contains an anchor element to the existing abstract section', () => {
 		initReadingCompanion();
 		const sections = readingCompanionContainer.querySelector('.c-reading-companion__sections-list');
 
@@ -205,7 +205,7 @@ describe('Reading Companion', () => {
 		expect(document.querySelector('.c-reading-companion__sections-list').firstElementChild.textContent).toBe('Abstract');
 	});
 
-	test('Should create list of items with existent figures', () => {
+	test('Should create a list of items with existent figures', () => {
 		includeMockFigures();
 		initReadingCompanion();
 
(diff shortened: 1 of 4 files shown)


toolkits/global/packages/global-article/__tests__/reading-companion.js
```diff
@@ -0,0 +1,394 @@
...
+		const referencesTab = document.querySelector('#tab-references');
+		const selectedReference = document.querySelector('#rc-ref-CR1');
+
+		selectedReference.scrollIntoView = jest.fn();
+		selectedReference.addEventListener = jest.fn();
+		emitter.emit('nav.reference', 'ref-CR1', null);
+
+		expect(referencesTab.getAttribute('aria-selected')).toBe('true');
+	});
+});
```
jon-stevens: I'm wondering if because the test file is fairly big, I would quite like to pull out the HTML into separate files, and maybe some of the functions that are only called once, could be put inline..

saramorueta: the file tested is actually big as well. We're planning to split it out in separate things as it is doing too many things.

You're right about adding that mocked HTML inline once and avoid calling those mock functions again and again. I actually thought to do it but I was a bit sick of this component and wanted to finish it haha! :stuck_out_tongue_closed_eye but yeah! you're right. Let's do the right thing :)

toolkits/global/packages/global-article/js/reading-companion.js
```diff
@@ -1,522 +1,527 @@
...
+
+		return {
+			init: function (config, scheduler, emitter) {
+				_access = Boolean(config.access);
+				_offset = config.offset || 0;
+
+				var initialised = false;
+
+				var setup = function (mql) {
+					if (initialised || mql.matches) { // eslint-disable-line no-use-before-define
```
moddular: Super minor thing, is this eslint comment still needed? It looks like you have got initialised defined already.

toolkits/global/packages/global-article/js/reading-companion.js
```diff
@@ -1,522 +1,527 @@
...
+					if (tabCount === 0) {
+						return;
+					}
+
+					bindEvents(scheduler, emitter);
+					if (tabCount > 1) {
+						switchToTab('sections');
+					}
+
+					initialised = true; // eslint-disable-line no-use-before-define
```
moddular: Same thing here - can we remove the eslint comment?

saramorueta:
Ok. I've done some refactoring in the tests. I haven't externalised the mock functions, but as @jon-stevens suggested, I moved some elements inline and kept only those functions for the changing cases.

Then I divided the all set of tests in two different blocks/levels which will differentiate when there are figures and references from when there won't.

I hope you find this more clear now.

Screen Shot 2020-03-23 at 09 01 46

toolkits/global/packages/global-article/js/reading-companion.js
saramorueta: good spot! thanks!

toolkits/global/packages/global-article/js/reading-companion.js
saramorueta: done!

saramorueta:
actually I slightly modified this to become this:

Screen Shot 2020-03-23 at 09 40 54

PR #211, opened 2020-03-23
initial attempt

matteo-gildone:
added new icon


matteo-gildone:
we don't need this anymore

PR #214, opened 2020-03-25
Global Expander: Updates close if click off option to include tab

benjclark:
It was strange that you could set the CLOSE_ON_CLICKOFF option to to false but yet when you tab out of the target it would close.

It makes more sense to tie the tabbing out and clicking out behaviour together under the same option therefore it now does so and it is called CLOSE_IF_NOT_FOCUSED which a better description of what it does.


toolkits/global/packages/global-expander/README.md
```diff
@@ -35,7 +35,7 @@ myExpander.init();
 | TARGET_HIDE_CLASS  | 'u-js-hide'   | String  | HTML class to be toggled on the target                                                                                |
 | TRIGGER_OPEN_CLASS | -             | String  | HTML class to be toggled to the trigger                                                                               |
 | TRIGGER_OPEN_LABEL | -             | String  | Text to set on the trigger when open                                                                                  |
-| CLOSE_ON_CLICKOFF  | true          | Boolean | Closes when you click outside of the target                                                                           |
+| CLOSE_IF_NOT_FOCUSED  | true          | Boolean | Closes when you click or tab outside of the target                                                                           |
```
davidpauljunior: I don't think the name needs to change personally, I think the clickoff is clearer as to what the option does / what it's intention is. If it was to change I think `CLOSE_ON_BLUR` or `CLOSE_ON_FOCUS_OUT` are better than having the IF in the name

davidpauljunior: If the name does change, this is a breaking change.

davidpauljunior: https://github.com/springernature/frontend-toolkits/pull/214/files#diff-80b465df5c418c066bb73b6f28bdb00dR46 would also need changing 😄

benjclark: ok, happy with `CLOSE_ON_FOCUS_OUT`

toolkits/global/packages/global-expander/__tests__/expander.spec.js
```diff
@@ -328,12 +328,12 @@ describe('Expander', () => {
 			expect(element.BUTTON.textContent).toEqual(triggerOpenLabelOption);
 		});
 
-		test('Should use CLOSE_ON_CLICKOFF option if it is passed to constructor', () => {
+		test('Should close target if target is clocked off if CLOSE_IF_NOT_FOCUSED option used', () => {
```
davidpauljunior: Why change the format of this test title? Compared to all the others which use the `Should use XXX_XXX option if it is passed to constructor`

benjclark: I just found it more descriptive as to what the test is trying to achieve and there's no having to imply or insinuate as much. Happy to keep it like the others though

davidpauljunior: The behaviour is actually tested earlier (`test('Should close when click off the target', () => {`. I wanted initially to just make an assertion that the correct option was passed to the constructor but it didn't assert that the option changed something. So I ended up with this, check the main behaviour, then check the option.

You are right that it doesn't describe what it does, but in that case we should change all the option based test titles.

benjclark: Ah ok, I hadn't seen the other test. Maybe this test isn't needed then as the other test actually already tests that the option is passed to the constructor?

davidpauljunior: Yeah, problem is that in the case of focus out, we initially test that it closes when you focus out because the option is true by default. But it doesn't test that it doesn't close if the option is false.
Perhaps the tests should be reorganised a bit. I tried to group them logically but it's not quite right.

davidpauljunior: So we could do:

```
test('Should close when focus out of the target', () => { ... });
test('Should NOT close when focus out of the target if CLOSE_ON_FOCUS_OUT is false', () => { ... });
```

And reorganised the others that have options similarly.

davidpauljunior: Sounds like a separate PR to me?

benjclark: I think they're fine tbh, unless you have a particular structure you want in mind?

davidpauljunior: Not urgently 👍

davidpauljunior:
So the only outstanding thing is whether or not this is a breaking change.

benjclark:
Hmm, yeah good point. I guess it is breaking for anyone that wanted tab out to close the popup when the old option CLOSE_ON_CLICKOFF option was disabled, however unlikely that actually is in reality.

davidpauljunior:
Very true. Carry on with the patch.

benjclark:
Ok, will do. Will leave it patch

PR #215, opened 2020-03-25
Buttons for springer toolkit

ricky-lau:
Add springer branded settings, button theme and responsive fonts.


toolkits/springer/packages/springer-buttons/HISTORY.md
```diff
@@ -0,0 +1,4 @@
+# History
+
+## 0.0.1 (2020-03-25)
```
davidpauljunior: `0.1.0` would match the first version of button?

toolkits/springer/packages/springer-buttons/scss/10-settings/buttons.scss
```diff
@@ -0,0 +1,78 @@
+/**
```
davidpauljunior: This file should be `button.scss` (now you've renamed global)

toolkits/springer/packages/springer-buttons/scss/50-components/buttons.scss
```diff
@@ -0,0 +1,15 @@
+.c-button {
+	@include font-size($button--font-size-normal)
```
davidpauljunior: I don't see any value in using this `font-size()` mixin. Especially for small things like button.

ricky-lau: I'm happy to remove it - i was imagining it might be more apparent when next to other links that do size down but being a button it'll probably be ok.

toolkits/springer/packages/springer-buttons/scss/10-settings/buttons.scss
ricky-lau: I noticed some packages have suffix `-override` - is that something to keep consistency with?

davidpauljunior: I started doing that initially, but it's actually an incorrect word for what is happening, as it could be extending rather than overriding.

davidpauljunior: So we opted for the brand to be a suffix. But haven't retrospectively fix the others.

toolkits/springer/packages/springer-buttons/scss/50-components/buttons.scss
davidpauljunior: I think use of that mixin should be at app level anyway tbh

sky-jack:
Yeah ditto. Will use this in my project in the near future.

PR #224, opened 2020-04-01
1st pass at show more

nickcall:
This is a first attempt and is here to get some feedback.


toolkits/global/packages/global-article/js/show-more.js
```diff
@@ -0,0 +1,116 @@
...
+
+			function setTrackingFor($button, state) {
+				if (settings.trackAttribute) {
+					$button.attr('data-track-action', settings.trackAttribute + ' show ' + state);
+				}
+			}
+
+			if (hasButton) {
+				$element.on('click', 'button[' + settings.linkAttribute + ']', function () {
+					if ($element.hasData(settings.expandedAttr)) {
```
nickcall, line 64: This is the line that is giving me the `TypeError: $element.hasData is not a function` when running the `Should toggle button between show more and show less` test

PR #225, opened 2020-04-02
Update global button

ricky-lau:
add padding to `.c-button--large`


davidpauljunior:
I'd say this is a patch

PR #228, opened 2020-04-07
add icon spacing to global-button

ricky-lau:
* BREAKING: Rename .c-button--block to .c-button--full-width
* BREAKING: Replace text-align with justify-content and variable rename
* Replace inline-block with inline-flex
* Add SVG spacing to buttons with .c-button-icon--left and .c-button-icon--right
* Move border-radius from base to themes
* Update readme

### DIFF
--- a/toolkits/global/packages/global-button/README.md
+++ b/toolkits/global/packages/global-button/README.md
@@ -29,9 +29,25 @@ Button styles to use on buttons and links.
 <button class="c-button c-button--primary">text</button>
 ```
 
-**Block**
+**Full width**
 ```html
-<button class="c-button c-button--block">text</button>
+<button class="c-button c-button--full-width">text</button>
+```
+
+**Icon left**
+```html
+<button class="c-button c-button--icon-left">
+    <svg></svg>
+    <span>text</span>
+</button>
+```
+
+**Icon right**
+```html
+<button class="c-button c-button--icon-right">
+    <span>text</span>
+    <svg></svg>
+</button>
 ```
 
 **Disabled**
--- a/toolkits/global/packages/global-button/scss/30-mixins/button.scss
+++ b/toolkits/global/packages/global-button/scss/30-mixins/button.scss
@@ -2,8 +2,9 @@
 // No theme
 
 @mixin button-base {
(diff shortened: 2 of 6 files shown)


toolkits/global/packages/global-button/README.md
```diff
@@ -29,9 +29,25 @@ Button styles to use on buttons and links.
 <button class="c-button c-button--primary">text</button>
 ```
 
-**Block**
+**Full width**
```
davidpauljunior, line 32: Any reason for this one?

toolkits/global/packages/global-button/scss/30-mixins/button.scss
```diff
@@ -2,18 +2,18 @@
 // No theme
 
 @mixin button-base {
+	align-items: center;
 	cursor: pointer;
-	display: inline-block;
+	display: inline-flex;
```
davidpauljunior, line 7: Should they not only be flexed for the `--icon` modifiers?

toolkits/global/packages/global-button/README.md
ricky-lau: I thought it might be potentially confusing to call it block when it has `display: flex` on it.

Also changed this to flex so that it can be used with the icon aligning classes. So that you can use the icon modifier without it being full width.

toolkits/global/packages/global-button/scss/30-mixins/button.scss
ricky-lau: Wanted to keep them all using flexbox so that it no longer needed text-align to do the same thing as justify-content (to keep it center by default).

Also reduced duplication of justify-content and align-items in both `--icon-right` and `--icon-left` as they can live in button-base mixin.

toolkits/global/packages/global-button/README.md
davidpauljunior: 👍

toolkits/global/packages/global-button/scss/30-mixins/button.scss
davidpauljunior: 👍

PR #229, opened 2020-04-08
Update breakpoints to use global

nickcall:
Making the article use the new global breakpoints so we can get consistency across the pages. Removes the need for article specific breakpoints hence the removal of $article--breakpoints

### DIFF
--- a/toolkits/global/packages/global-article/scss/50-components/article-enhanced.scss
+++ b/toolkits/global/packages/global-article/scss/50-components/article-enhanced.scss
@@ -7,7 +7,7 @@
 	width: 60.2%;
 	margin-right: 8.6%;
 
-	@include media-query(map-get($article--breakpoints, 'med'), 'max') {
+	@include media-query('md', 'max') {
 		width: 100%;
 		margin-right: 0;
 	}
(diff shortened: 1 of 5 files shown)


toolkits/global/packages/global-article/scss/50-components/article-enhanced.scss
```diff
@@ -386,7 +386,7 @@
 	padding: 0;
 }
 
-@include media-query(map-get($article--breakpoints, 'lrg') map-get($article--breakpoints, 'med'), 'range') {
+@include media-query('lg' 'md', 'range') {
```
davidpauljunior, line 389: Are these not backwards? From md - lg, rather than lg - md. No idea what difference it makes, if any.

nickcall: This appears to be correct, at least not changed from the original.

```@media only screen and (min-width: 1024px) and (max-width:767px) {
    .c-article-section__illustration-center,.c-article-section__illustration-right {
        margin:0 auto 24px;
        float: none;
        max-width: 100%
    }
}```

PR #245, opened 2020-04-23
Allow for explorers lovely non-standard key identifiers

jon-stevens:
Internet Explorer uses non-standard key identifiers, so this change allows for 'Down' in place of 'ArrrowDown' and 'Up' in place of 'ArrowUp'.

### DIFF
--- a/toolkits/global/packages/global-autocomplete/js/index.js
+++ b/toolkits/global/packages/global-autocomplete/js/index.js
@@ -68,7 +68,7 @@ const autoComplete = arguments_ => {
 		}
 
 		container().addEventListener('keydown', event => {
-			if (['Escape', 'ArrowUp', 'ArrowDown'].includes(event.key)) {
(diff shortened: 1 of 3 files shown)


toolkits/global/packages/global-autocomplete/js/index.js
```diff
@@ -78,40 +78,33 @@ const autoComplete = arguments_ => {
...
+					}
+					previousSibling.focus();
+				} else {
+					input.focus();
+					if (selectOnSuggestionBrowsing) {
+						input.value = currentSearchTerm;
+					}
+				}
+			}
+			else if (event.key === 'Escape') {
```
morgaan, line 105:
```suggestion
			} else if (event.key === 'Escape') {
```

Another lint issue.

Almost complained about those one, but I thought this passed through the linter already, and that you were leveraging the syntax for readability sake. Sorry.

jon-stevens: yeah I should have run the lint script before committing - always forget!
I'm used to having lint and test run together..

PR #249, opened 2020-05-04
Move context

alexkilgour:
Combine the existing context packages to create a new `brand-context`
- `global-context`
- `nature-context`
- `springer-context`
- `springer-nature-context`

The new `brand-context` package exports configurations for the different brands. The brand names map to the toolkit names, except in the case of global which uses the new `default` brand (global-context configuration has been moved to the default brand in brand-context).

The other change is that _all_ the imprint brands - nature, springer, springernature - include the default brand first, then any brand specific configuration overwrites/adds to those that come from default.

This update uses the new version of the `frontend-package-manager` which removes the extended package functionality. This no longer happens on package publication.

## Consequences

As a result of merging this PR, changes will no longer be able to be made to the `context` until components have been updated to use the new `brand-context`, and updates will no longer be able to be made to extended packages until they are updated to global branded packages.


context/brand-context/README.md
```diff
@@ -0,0 +1,35 @@
+# Brand Context
+
+The `brand-context` package contains baseline configurations that provide skaffolding for products and toolkit components within each of the supported brands:
```
davidpauljunior, line 3:
```suggestion
The `brand-context` package contains baseline configurations that provide scaffolding for products and toolkit components within each of the supported brands:
```

context/brand-context/default/README.md
```diff
@@ -0,0 +1,175 @@
...
+- [Colors](#colors)
+	- [Greyscale](#greyscale)
+	- [Color](#color)
+	- [Foreground color](#foreground-color)
+- [Layers](#layers)
+- [Icons](#icons)
+
+## Breakpoints
+
+The `media-query` @mixin is provided for handling breakpoints. Breakpoint values are provided at the `10-settings` level using the `$breakpoints` map.
```
davidpauljunior, line 15:
```suggestion
The `media-query` mixin is provided for handling breakpoints. Breakpoint values are provided at the `10-settings` level using the `$breakpoints` map.
```

PR #280, opened 2020-06-04
Sn button context

sky-jack:
Includes configuration for brand-specific button styles in Springer Nature brand context.

### Updates @alexkilgour 

Needs discussion with @sky-jack 

Removed `disabled` button config for SpringerNature as they don't have one yet, this means it falls back to the one defined in the default brand. Also, as seen in the screenshot below, `ghost` is not defined in either the SN or default brands, so it falls back to the `default` button style.

The following styles are not defined in the SN brand so come from the default brand. These differ from those in [Anura](https://github.com/springernature/anura/blob/master/anura-lipstick/resources/scss/40-components/buttons.scss) and probably Identity. This is probably OK to commit now, but I think needs alignment across SN branded products then updating

```
$button--border-width-style: 1px solid;
$button--line-height: 1.3;
$button--font-size-small: 1.4rem;
$button--font-size-large: 1.8rem;
$button--icon-margin: spacing(8);
$button--justify-content: center;
$button--transition: 0.25s ease, color 0.25s ease, border-color 0.25s ease;
```

Screenshot 2020-06-05 at 09 46 09


context/brand-context/default/scss/10-settings/buttons.scss
```diff
@@ -19,6 +19,8 @@ $button--padding: spacing(8);
 $button--padding-large: spacing(16);
 $button--justify-content: center;
 $button--transition: 0.25s ease, color 0.25s ease, border-color 0.25s ease;
+$button--border-radius: null;
```
alexkilgour, line 22: these are specific to `springernature`, setting them to null be default means they won't be used for other brands

sky-jack:
This looks good to me, I will review my brand level settings before I integrate.

PR #285, opened 2020-06-15
Tweaks springer header background colour and bottom border colour

benjclark:
as per tweaks requested by James A in https://jira.springernature.com/browse/PAN-755

### DIFF
--- a/toolkits/springer/packages/springer-header/scss/50-components/_header.scss
+++ b/toolkits/springer/packages/springer-header/scss/50-components/_header.scss
@@ -3,7 +3,7 @@
 }
 
 .c-header {
-	background-color: #fff;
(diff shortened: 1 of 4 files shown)


toolkits/springer/packages/springer-header/scss/50-components/_header.scss
```diff
@@ -3,9 +3,9 @@
 }
 
 .c-header {
-	background: $header--background;
+	background-color: #fff;
```
davidpauljunior, line 6: Why not change the value in the settings?

benjclark: Oh yeah derrrr, for some reason that didn't even occur

PR #289, opened 2020-06-29
Adds option to expander for callback that fires just before focus and…

benjclark:
… adds tests

If you want to use global-expander in a component that also needs to add click event handlers to the expander trigger element then you might run in to race condition issues between the handler that expander adds and the handler that your component adds.

This adds an option to expander to allow you to pass in a callback that is called just before expander focuses on the target element. And therefore prevents consumers of the expander from having to manage sequencing of handlers / race conditions.


benjclark:
Spoke to ajk about this and going to refactor this to work off a custom event mechanism opposed to a callback

PR #291, opened 2020-06-30
Makes it clear that open and close method are public

benjclark:
Whether it was intentional or not I'm unsure but it already was the case that you can call `expander._close()` and `expander._open()` but this makes it more obvious when reading the code.

Also note that it is needed to expose these methods. For example if you have a close button on a popup that uses global-expander. That close button needs a way to close the popup, the best way being to manually call something like `expander.close()` (because you need to update expander instance's `isOpen` state)

### DIFF
--- a/toolkits/global/packages/global-expander/js/expander.js
+++ b/toolkits/global/packages/global-expander/js/expander.js
@@ -39,7 +39,7 @@ const Expander = class {
 		event.preventDefault();
 
 		if (this._isOpen) {
-			this._close();
+			this.close();
 		} else {
 			this.open();
 		}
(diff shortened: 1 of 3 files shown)


toolkits/global/packages/global-expander/js/expander.js
```diff
@@ -41,7 +41,7 @@ const Expander = class {
 		if (this._isOpen) {
 			this._close();
```
alexkilgour, line 42: should this not be `close()` now (and a bit further down)?

benjclark: Whoops. I thought I'd don't a find and replace but clearly hadn't. Thanks

PR #292, opened 2020-07-01
initial setup

matteo-gildone:
Breadcrumb component like on Journal pages.

Screen Shot 2020-07-01 at 08 35 26


toolkits/global/packages/global-breadcrumb/scss/10-settings/default.scss
```diff
@@ -0,0 +1,6 @@
+$breadcrumb-font-family: $font-family-sans;
+$breadcrumb-font-size-sm: $font-size-sm;
```
alexkilgour, line 2: `$font-size-sm` is from the springer context and won't be available here. I would just whatever pixel value this represents here.

toolkits/global/packages/global-breadcrumb/scss/50-components/breadcrumb.scss
```diff
@@ -0,0 +1,14 @@
+.c-breadcrumb {
+	@include u-list-inline;
+	@include font-size($breadcrumb-font-size-sm);
```
alexkilgour, line 3: `font-size()` mixin is Springer only at the moment, just reference the variable

PR #294, opened 2020-07-01
rename component and add nature font size variation

matteo-gildone:


toolkits/global/packages/global-breadcrumbs/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@springernature/global-breadcrumbs",
-    "version": "0.0.1",
+    "version": "1.0.0",
```
alexkilgour, line 3: this should be `0.1.0` to match history

PR #297, opened 2020-07-03
Move RFS to be internal to brand-context

alexkilgour:
Springer uses [`RFS`](https://github.com/twbs/rfs/tree/v8.0.4) as an external dependency, but because of the way the package is written we cannot include it in the `brand-context` whilst maintaining our desired source ordering. This has meant that apps have to manually include (1) the `RFS` settings from `brand-context`, then (2) the `RFS` library itself at the top of the main include.

This has caused issues for apps that don't realise it needs to be included, and it would be much easier if we could bundle this within `brand-context`.

This PR takes the `RFS` code and embeds it within `brand-context` so that we no longer require a dependency. It makes the following changes:

1. Splits the `RFS` code into `settings` and `mixins` to fit within our architecture
2. Removes the use of `!default` so that we can override settings on a per brand basis
3. Has to duplicate the caching inside each brand (currently only springer) settings file so the correct values are generated
4. Add SASS testing (via new utility) to verify that the parts of `rfs` we use are working as expected

This now means that `RFS` is available to _all_ brands, not just Springer. There is a question as to whether we use this at all going forward, but for now, we do and this should make adoption of the `brand-context` easier for developers.


benjclark:
Has all this code been linted? Looks like some files have 2 spaces indents and others use tab indents. Tab indents seems to be what js and scss files use across toolkits from my quick check

alexkilgour:
> Has all this code been linted? Looks like some files have 2 spaces indents and others use tab indents. Tab indents seems to be what js and scss files use across toolkits from my quick check

Hmmm, it passes all the linting checks, let me look

context/brand-context/default/scss/10-settings/rfs.scss
```diff
@@ -0,0 +1,69 @@
...
+
+@if $rfs-breakpoint-unit != px and $rfs-breakpoint-unit != em and $rfs-breakpoint-unit != rem {
+  @error "`#{$rfs-breakpoint-unit}` is not a valid unit for $rfs-breakpoint-unit. Use `px`, `em` or `rem`.";
+}
+
+// Resize font size based on screen height and width
+$rfs-two-dimensional: false;
+
+// Factor of decrease
+$rfs-factor: 10;
```
benjclark, line 31: springer currently sets this as `1.6` as I'm guessing `10` was too extreme. Was this the value copied directly from their source code or did you decide upon this as the default?

context/brand-context/default/scss/10-settings/rfs.scss
```diff
@@ -0,0 +1,69 @@
...
+
+@if type-of($rfs-factor) != "number" or $rfs-factor <= 1 {
+  @error "`#{$rfs-factor}` is not a valid  $rfs-factor, it must be greater than 1.";
+}
+
+// Generate enable or disable classes. Possibilities: false, "enable" or "disable"
+$rfs-class: false;
+
+// 1 rem = $rfs-rem-value px
+$rfs-rem-value: 16;
```
benjclark, line 41: brand context's `typography.scss` currently sets the font-size as `62.5%` on the html element which in effect means `font-size: 10px`  as the baseline font size on our sites. Therefore we set this variable to `10` in our rfs setup on springer to tell the plugin that.

context/brand-context/default/scss/10-settings/rfs.scss
alexkilgour: This is in the `default` brand, and is an exact copy of the `rfs` defaults. There is a `springer/scss/10-settings/rfs.scss` file that sets the springer settings to override these

context/brand-context/default/scss/10-settings/rfs.scss
benjclark: https://github.com/twbs/rfs#rem-value-number for reference

alexkilgour: this is overridden in `springer/scss/10-settings/rfs.scss`

benjclark: if we are going to have 62.5% set as the default for projects using brand context maybe this default should align to prevent every project from having to override it?

alexkilgour: the settings in `default` are all exactly as `rfs` defined them

context/brand-context/default/scss/10-settings/rfs.scss
benjclark: oh yeah I just meant maybe we should default to a low factor and let people choose more extreme factors if they have a specific need

context/brand-context/default/scss/10-settings/rfs.scss
benjclark: oh yeah I get that, I'm just saying that we could save everyone the hassle of having to override this by changing this default to `10` which would align with the decision we made in brand context to default the font-size to `62.5%`. Otherwise every project that uses brand-context (which I assume all will?) will need to know that they need to override the `$rfs-rem-value` to `10` in their rfs settings file

alexkilgour: Yeah, that does make sense. I was taking the approach of exactly replicating `rfs` so the process was the same, but maybe this does make things easier

alexkilgour:
I have updated the `default` font-size to `10px` (the same as springer) so that `html { font-size: 62.5% }` works for the default brand. Have changed the `rfs-factor` to `1.2`, more sensible than `10` but still different from springer as one of the important things to test is that the imports work and brands can override `default` settings. Having at least one difference means that I can test this!

PR #299, opened 2020-07-06
Add aria-pressed to support non-native buttons on expander

ricky-lau:


benjclark:
When you come to bump the version in Oscar can you make sure you bump the global-popup to use this version of global-expander too. Thanks!

PR #300, opened 2020-07-07
Setup for an a/b test to make all fonts sans-serif on some journals a…

moddular:
…nd trim down the serif font stack

### DIFF
--- a/context/brand-context/HISTORY.md
+++ b/context/brand-context/HISTORY.md
@@ -1,11 +1,14 @@
 # History
 
+## 3.2.4 (2020-07-07)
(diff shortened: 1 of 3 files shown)


benjclark:
How does this work in the context of an AB test then? Those are variables in scss which are only of importance at build time? How does that translate in to an AB test?

moddular:
The A/B test groups are served different stylesheets. One of these sets the $is-sans variable to true the other doesn't.

moddular:
Hopefully this won't need to be around for long so the 2 stylesheets can be merged back into one again once the test is over

PR #301, opened 2020-07-10
Adds getCookie, Debounce and Throttle utils to global-javascript

tulipdexter:


benjclark:
I remember making debounce and throttle functions a few years ago at SN and at the time the advice was to use requestAnimationFrame instead of setTimeout because of performance issues. Can't remember if it was for both of them or one of them

benjclark:
https://gomakethings.com/debouncing-your-javascript-events/

https://www.html5rocks.com/en/tutorials/speed/animations/

benjclark:
https://stackoverflow.com/questions/41740082/scroll-events-requestanimationframe-vs-requestidlecallback-vs-passive-event-lis

Kind of depends what you are using the debounce / throttle for. Might be safest to use rAF seeing as this is a generic helper that anyone can use for any purpose?

tulipdexter:
That's really interesting.
The only thing with rAF is that is removes the applications control of the execution rate (rAF will aim for about 16ms).
For things like autocomplete, where we're making ajax request, that seems a little extreme.

I wonder then if the timer should default to 16, and if it is 16 or less it uses rAF, anything above will setTimeout and take the wait time passed in.

What do you think?

benjclark:
I don't think 16ms is anything worth worrying about to be honest. And with something like an autocomplete I guess you'd want to wait until the browser is ready to do it's next repaint before displaying anything, to be efficient?

tulipdexter:
Yeah but if someone is typing and you need to make an ajax request based on what they type, you probably don't want to be making that call to the endpoint every 16ms. More likely 200ms or something?

benjclark:
Yeah true. Your suggestion is best then, nice

toolkits/global/packages/global-javascript/README.md
```diff
@@ -64,6 +67,42 @@ elementToBind.addEventListener('componentName:eventName', function (event) {
...
 
+#### getCookie
+Retrieves a cookie by name from `document.cookie`.
+
+```javascript
+const myCookie = getCookie('name-of-cookie');
+```
+
+#### debounce
+Allows sequential calls to function to be grouped together so that the function will only be called once.
```
benjclark, line 78:
```suggestion
Allows sequential calls to a function to be grouped together so that the function will only be called once.
```

toolkits/global/packages/global-javascript/src/helpers/util/debounce.js
```diff
@@ -0,0 +1,46 @@
...
+			}
+		};
+		const callNow = immediate && !timeout;
+
+		// If there's a timer, cancel it
+		if (timeout) {
+			if (raf) {
+				window.cancelAnimationFrame(timeout);
+			} else {
+				console.log('into clear timeout');
```
benjclark, line 35: bye bye

toolkits/global/packages/global-javascript/__tests__/unit/util/throttle.spec.js
```diff
@@ -0,0 +1,37 @@
...
+beforeEach(() => jest.useFakeTimers());
+
+describe('throttle', () => {
+	test('Should only call func once within the wait time', () => {
+		const func = jest.fn();
+		const throttledFunc = throttle(func, 500);
+
+		// Call it several times with 100ms between each call
+		for(let i = 0; i < 5; i++) {
+			jest.advanceTimersByTime(100);
```
benjclark, line 12: this is cool

PR #302, opened 2020-07-10
Styles for Global-Popup and refactors redundant code

benjclark:
- Adds basic popup styling to Global Popup
- Removes basic popup styling from Global Article enhanced-article.scss
- Imports and uses Global-Popup in Global Article authors.js
- Makes Global-Popup and Global-Expander dependencies of Global-Article (because of authors.js)
- Bumps Global Popup to v0.2.0 and Global Article to v20.0.0

Follow up to this is https://github.com/springernature/oscar/pull/1969


toolkits/global/packages/global-article/__tests__/authors.js
```diff
@@ -22,13 +20,13 @@ describe('Authors', () => {
 		let listChildren = '';
 
 		for (let i = 0; i < number; i++) {
-			listChildren += `<li><a href="#author${i}">author${i}</a><sup><a href="#Aff${i}" class="js-no-scroll" tabindex="-1">1</a>
-				${separator ? `<span class="js-separator">,</span>` : ''}
+			listChildren += `<li><a href="#author${i}">author${i}</a><sup><a href="#Aff${i}" tabindex="-1" data-author-popup="author${i}">1</a>
```
tulipdexter, line 23: I think this would be cleaner if it was creating the element and appending it to a fragment rather than creating a string. Innerhtml is pretty slow I think

toolkits/global/packages/global-article/__tests__/authors.js
```diff
@@ -65,26 +63,26 @@ describe('Authors', () => {
...
 		});
-	
+
 		test('Should register click events for each author name', () => {
-			jest.spyOn($.fn, 'delegate');
-	
-			authorList.create($(selector));
-	
-			expect($('.c-author-list').delegate).toHaveBeenCalled();
+			selector.addEventListener = jest.fn();
```
tulipdexter, line 81: This doesn't really check that it registers click events, just that the eventListener method was called

toolkits/global/packages/global-popup/scss/10-settings/_global-popup.scss
```diff
@@ -0,0 +1,6 @@
+$c-popup--padding-vertical: spacing(16);
```
tulipdexter, line 1: These variables feel a bit odd to me.
Vertical is generic for top and bottom, but left and right are different.
Would `$c-popup--padding: spacing(16, 32, 16, 24)` make more sense?

I know it's used later on to determine the position of the icon, but one of them is divided by two anyway so it's not manipulating the number. Maybe vars like `$c-popup--close-top` would make it clearer?

toolkits/global/packages/global-popup/scss/50-components/_global-popup.scss
```diff
@@ -0,0 +1,38 @@
+.c-popup {
+	position: absolute;
+	font-family: $font-family-serif;
+	z-index: 100;
+	padding: $c-popup--padding-vertical $c-popup--padding-right $c-popup--padding-vertical $c-popup--padding-left;
+	border: 1px solid $c-popup--border-color;
+	box-shadow: $c-popup--box-shadow;
+	min-width: 300px;
```
tulipdexter, line 8: Variables?

toolkits/global/packages/global-article/__tests__/authors.js
benjclark: Yeah tbh this entire file needs refactoring in to es6 and improving in many places. I'll do this though as a nice little tidy up step

benjclark: Oh wait, I thought this comment was about authors.js but it's about the test for authors.js. Did you mean to post it about the module authors.js? Because that's a legit comment about the use of `.innerHTML =` in that file. But I guess in terms of the test file it doesn't matter how efficiently the dom setup logic is, our laptops are fast enough that the difference with be in perceivable?

tulipdexter: Oh yeah sorry, my mistake. Didn't realise it was a test file

benjclark: But the funny thing is, you make a good point about the actual module! haha

PR #303, opened 2020-07-13
remove greyscale function

matteo-gildone:
$greyscale-min and $greyscale-max on springer are different from the default one, generating different link colours.
The colour of links should be consistent across the platforms.


tulipdexter:
Does this mean that those min max values in Springer should be changed?
https://github.com/springernature/frontend-toolkits/blob/master/context/brand-context/springer/scss/10-settings/colors/default.scss

matteo-gildone:
If we change or remove min max values from springer then we need to update all the greyscale usage on springer. That's why I asked in the channel if someone knew why they are different. Personally I think greyscale should behave the same way but there might be valid reasons to have different min max values.

PR #304, opened 2020-07-14
global-javascript: Adds OneTrust utilities

tulipdexter:
Adds `checkConsent` and `isConsentBannerClosed` utilities.

### DIFF
--- a/toolkits/global/packages/global-javascript/README.md
+++ b/toolkits/global/packages/global-javascript/README.md
@@ -18,8 +18,8 @@ You can import as many of the named exports from the helpers as you require for 
 - [makeArray](#makearray)
 - [createEvent](#createevent)
 - [getCookie](#getcookie)
-- [getCookie](#debounce)
-- [getCookie](#throttle)
+- [debounce](#debounce)
+- [throttle](#throttle)
 - [onetrust](#onetrust)
 
 
--- a/toolkits/global/packages/global-javascript/README.md
+++ b/toolkits/global/packages/global-javascript/README.md
@@ -20,10 +20,11 @@ You can import as many of the named exports from the helpers as you require for 
 - [getCookie](#getcookie)
 - [getCookie](#debounce)
 - [getCookie](#throttle)
+- [onetrust](#onetrust)
 
 
 **Dom**
-- [getDataOptions](#getDataOptions)
+- [getDataOptions](#getdataoptions)
 
 ### Util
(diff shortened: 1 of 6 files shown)


toolkits/global/packages/global-javascript/README.md
```diff
@@ -20,10 +20,11 @@ You can import as many of the named exports from the helpers as you require for
 - [getCookie](#getcookie)
 - [getCookie](#debounce)
 - [getCookie](#throttle)
```
alexkilgour, line 22: Not part of this PR, but the name are wrong on these!
```
[getCookie](#debounce)
[getCookie](#throttle)
```

Should be:

```
[debounce](#debounce)
[throttle](#throttle)
```

tulipdexter: 😬

toolkits/global/packages/global-javascript/src/helpers/util/onetrust.js
```diff
@@ -0,0 +1,36 @@
...
+	};
+
+	if (!validCategories[category]) {
+		throw new Error('Invalid category: ' + category);
+	}
+
+	const consent = getCookie('OptanonConsent');
+	const consentGroups = consent.split('groups=').pop().split('&')[0];
+	const colonEncoding = '%3A';
+
```
alexkilgour, line 25: Is it likely that any of this could change? either by us or onetrust. i.e. would it be valuable to make some of this configurable with an `init()` function, such as categories, cookie name? It might be that we hope not, and just cross that bridge when/if it happens, but it's easier for an app to update a config than release a new version of this package every time something changes.

tulipdexter: Hmm good question. I think that it's unlikely that the cookie name would change, but it's possible that the categories could change (based on our side). I could make the categories I've set here the default option and allow different categories to be set via an init.

alexkilgour: If that's easy then it would safeguard against them changing

alexkilgour: After discussion offline, we think it is best to leave it as it is. We want the module to reflect the current state of our OneTrust configuration, and making it configurable would mean those categories _could_ go stale. Every team should be using the same set of categories; if they change for one they change for all. So a new version will be released if the categories change, and teams will update.

PR #306, opened 2020-07-14
Fix space key and refactor AUTOFOCUS option on Global expander

ricky-lau:
* ~Fix bug with detecting 'Space' key~
* ~Add TARGET_FOCUS option to control focus moving to target~
* ~Add tests for 'Space' key on non-native button~
* ~Add tests for TARGET_FOCUS option~

Update:
* Refactor AUTOFOCUS from boolean to string to set options
* Update readme on refactored AUTOFOCUS option
* Fix bug with detecting 'Space' key
* Add tests for 'Space' key on trigger
* Add tests for anchor link button


toolkits/global/packages/global-expander/__tests__/expander.spec.js
```diff
@@ -19,7 +19,7 @@ const createKeydownEvent = key => {
 			event.key = 'Enter';
 			break;
 		case 'Space':
-			event.key = 'Space';
+			event.key = ' ';
```
tulipdexter, line 22: _googles_ 😆 
Nice catch. Have you checked this works in IE? I read a few things saying it needs 'Spacebar' 🤷‍♂️

ricky-lau: 😱 

Of course IE would do something different 😩

toolkits/global/packages/global-expander/js/expander.js
```diff
@@ -9,6 +9,7 @@ const defaultOptions = {
 	TRIGGER_OPEN_CLASS: 'is-open',
 	TRIGGER_OPEN_LABEL: undefined,
 	CLOSE_ON_FOCUS_OUT: true,
+	TARGET_FOCUS: true,
```
tulipdexter, line 12: Does this option make the `AUTOFOCUS` feel a bit confusing?
What happens if they're both set to true?

One focuses on the target element, the other focuses on the first tabbable element _inside_ the target.

tulipdexter: I'm wondering if you should keep `AUTOFOCUS` but it should take a string instead of a boolean.
Maybe `AUTOFOCUS: 'target'`, `AUTOFOCUS: 'child'`, `AUTOFOCUS: null` (null being the default 🤔 ?

toolkits/global/packages/global-expander/__tests__/expander.spec.js
ricky-lau: added 'Spacebar' to get it detected by older browsers that use this

toolkits/global/packages/global-expander/js/expander.js
```diff
@@ -152,18 +152,23 @@ const Expander = class {
 			this._triggerEl.dispatchEvent(event);
 		}
 
-		if (this._options.AUTOFOCUS) {
-			if (this._targetTabbableItems.length > 0) {
-				const firstTabbableItem = this._targetTabbableItems[0];
-				firstTabbableItem.focus();
+		switch (this._options.AUTOFOCUS) {
```
tulipdexter, line 155: I might be wrong, but this doesn't feel like it should be a switch statement to me, mainly because there's no default behaviour, it's just an if else if.

Further to that, should we be throwing an error if a string that we don't expect is passed in?

ricky-lau: I'll make new PR for this

PR #310, opened 2020-07-20
Updates close icon of Global Popup to icon from global icon set

benjclark:


toolkits/global/packages/global-popup/scss/50-components/_global-popup.scss
```diff
@@ -16,9 +16,9 @@
...
-	padding-right: 12px;
+	height: 16px;
+	background: url("data:image/svg+xml;base64,PHN2ZyBoZWlnaHQ9IjE2IiB2aWV3Qm94PSIwIDAgMTYgMTYiIHdpZHRoPSIxNiIgeG1sbnM9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIj48cGF0aCBkPSJtMi4yOTY3OTU3NSAxMi4yNzcyNDc4Yy0uMzk2NTg3NTcuMzk2NTg3Ni0uMzk0Mzg4NDcgMS4wMzI4MTA5LS4wMDA2MjE0OCAxLjQyNjU3NzkuMzk2NTEyMjcuMzk2NTEyMyAxLjAzMjQ2NzY4LjM5MzQ4ODggMS40MjY1Nzc5MS0uMDAwNjIxNGw0LjI3NzI0NzgyLTQuMjc3MjQ3ODcgNC4yNzcyNDc4IDQuMjc3MjQ3ODdjLjM5NjU4NzYuMzk2NTg3NSAxLjAzMjgxMDkuMzk0Mzg4NCAxLjQyNjU3NzkuMDAwNjIxNC4zOTY1MTIzLS4zOTY1MTIyLjM5MzQ4ODgtMS4wMzI0Njc3LS4wMDA2MjE0LTEuNDI2NTc3OWwtNC4yNzcyNDc4Ny00LjI3NzI0NzggNC4yNzcyNDc4Ny00LjI3NzI0NzgyYy4zOTY1ODc1LS4zOTY1ODc1Ny4zOTQzODg0LTEuMDMyODEwOTEuMDAwNjIxNC0xLjQyNjU3NzkxLS4zOTY1MTIyLS4zOTY1MTIyNi0xLjAzMjQ2NzctLjM5MzQ4ODc1LTEuNDI2NTc3OS4wMDA2MjE0OGwtNC4yNzcyNDc4IDQuMjc3MjQ3ODItNC4yNzcyNDc4Mi00LjI3NzI0NzgyYy0uMzk2NTg3NTctLjM5NjU4NzU3LTEuMDMyODEwOTEtLjM5NDM4ODQ3LTEuNDI2NTc3OTEtLjAwMDYyMTQ4LS4zOTY1MTIyNi4zOTY1MTIyNy0uMzkzNDg4NzUgMS4wMzI0Njc2OC4wMDA2MjE0OCAxLjQyNjU3NzkxbDQuMjc3MjQ3ODIgNC4yNzcyNDc4MnoiIGZpbGw9IiM2NjYiIGZpbGwtcnVsZT0iZXZlbm9kZCIvPjwvc3ZnPg==") no-repeat;
```
tulipdexter, line 20: This will need updating if the global icons change, and that will only happen if someone happens to know it's changed.
It would be great if close icons could be done using the html character `&times;`. Is it worth investigating to remove the need for a background image?

benjclark: James A said it was the rounded corners of the close X that he wanted, which is what the the icon in the global icon set has. Do you think rounded corners can be achieved if we use `&times` ?

saramorueta, line 20: I'm just wondering if the encoding is correct 🤔

benjclark: Yeah I've tested it and it works. Maybe it's because I base64 encoded it while the previous one wasn't base64?

benjclark: Good spot though. I've removed the base64 encoding, now it's just svg inline and it's less code overall which is good

tulipdexter: No I doubt it. I think there's something missing with using icons from the global set in components. If copying and pasting is the only way then fair enough, just feels a bit fragile

saramorueta: well, I remember that a long time ago we modified all the icons on the article pages to be `utf-8`but I don't remember exactly if it was for some incompatibility issue or for some performance tweaks. In any case, if it works I guess It's fine! I just wanted to double-check!

benjclark: Yeah I guess what would be good in future is for us to implement some kind of placeholder replacer thing in toolkit or oscar scss for inserting global icons in css

saramorueta: oh, you've changed it! haha ok!

PR #313, opened 2020-07-20
handle AUTOFOCUS option in global-expander

ricky-lau:
- handle values that are not accepted for AUTOFOCUS
- swap out switch for if/else when detecting values on AUTOFOCUS

### DIFF
--- a/toolkits/global/packages/global-expander/js/expander.js
+++ b/toolkits/global/packages/global-expander/js/expander.js
@@ -25,10 +25,14 @@ const Expander = class {
 
 		this._isOpen = false;
 
-		this._handleButtonClick = this._handleButtonClick.bind(this);
-		this._handleButtonKeydown = this._handleButtonKeydown.bind(this);
-		this._handleDocumentClick = this._handleDocumentClick.bind(this);
-		this._handleDocumentKeydown = this._handleDocumentKeydown.bind(this);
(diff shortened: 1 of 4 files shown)


toolkits/global/packages/global-expander/js/expander.js
```diff
@@ -123,6 +123,25 @@ const Expander = class {
...
 
+	/**
+	 * AutoFocus
+	 */
+
+	_handleAutoFocus() {
+		if (this._options.AUTOFOCUS === 'target') {
+			this._autoFocusElement = this._targetEl;
+			this._targetEl.setAttribute('tabindex', '-1');
+			this._autoFocusElement.focus();
```
tulipdexter, line 134: This can happen at the end of this function.

toolkits/global/packages/global-expander/js/expander.js
```diff
@@ -123,6 +123,25 @@ const Expander = class {
...
+		}
+		if (this._options.AUTOFOCUS === 'firstTabbable') {
+			this._autoFocusElement = this._targetTabbableItems.length > 0 && this._targetTabbableItems[0];
+			if (this._autoFocusElement.setSelectionRange) {
+				this._autoFocusElement.setSelectionRange(0, this._autoFocusElement.value.length);
+			}
+			this._autoFocusElement.focus();
```
tulipdexter, line 141: This won't be needed if it's moved to the end of the function

PR #321, opened 2020-07-31
Removes coverage report from every test run

benjclark:
The coverage report always slows down tests and imo there's no need to run it every time, it would be better if we run it ad hoc?

### DIFF
--- a/context/brand-context/package.json
+++ b/context/brand-context/package.json
@@ -4,8 +4,5 @@
   "license": "MIT",
   "description": "Bootstrapping for all components and products",
   "keywords": [],
-  "author": "Springer Nature",
-  "devDependencies": {
-    "@springernature/util-sass-compiler": "^0.1.0"
-  }
+  "author": "Springer Nature"
 }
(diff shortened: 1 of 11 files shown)


jpw:
As for the first thing, have you run the bootstrap command? https://github.com/springernature/frontend-toolkits#npm-scripts

(I hit this recently in another repo, perhaps it is time to rethink this approach, but not sure what we can do...?)

context/brand-context/package.json
```diff
@@ -4,5 +4,8 @@
   "license": "MIT",
   "description": "Bootstrapping for all components and products",
   "keywords": [],
-  "author": "Springer Nature"
+  "author": "Springer Nature",
+  "devDependencies": {
```
tulipdexter, line 8: I'm a bit confused, what is this for?

benjclark: The Jest tests fail if the @springernature imports in the tests and actual modules are not installed in the package

benjclark:
@jpw  Ah yes, thank you. I hadn't spotted that before. That's good to know. Thanks!

PR #324, opened 2020-08-06
Removes hook option from global-popup

benjclark:
The hook option was causing issues on mobile so needed a rethink. It's better to always position a popup relative to the body. It makes `width: auto` easier to think about as well as any `MAX_WIDTH` option that may have been set.

### DIFF
--- a/toolkits/global/packages/global-popup/js/popup.js
+++ b/toolkits/global/packages/global-popup/js/popup.js
@@ -130,7 +130,7 @@ const Popup = class {
 		return {
 			left: (windowWidth < 600) ? 0 : offset.left,
 			top: top
-		}
+		};
 	}
 };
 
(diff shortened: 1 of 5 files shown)


toolkits/global/packages/global-popup/js/popup.js
```diff
@@ -110,8 +104,8 @@ const Popup = class {
 		// calc space above trigger
 		const spaceAbove = offset.top - this._content.offsetHeight - arrowHeight;
 
-		const abovePosition = 0 - this._content.offsetHeight - arrowHeight;
-		const belowPosition = 0 + triggerMetrics.height + arrowHeight;
+		const abovePosition = offset.top - this._content.offsetHeight - arrowHeight;
```
tulipdexter, line 107: 👋 I presume it's no coincidence that this is the same calculation as `spaceAbove`. Does it make the `spaceAbove` var redundant?

PR #332, opened 2020-08-24
global-javascript: Adds setCookie and deleteCookie utilities

tulipdexter:


benjclark (approved):
bono

jpw (approved):
Looks nice, I wonder if a nice addition might be to warn if the encoded cookie size is over the max size of 4096 byes...

tulipdexter:
@jpw go for it!

PR #337, opened 2020-09-08
Fix references to global-spacing in typography.scss

sonniesedge:
There are still some references to the `global-spacing` function, which does not exist and seems to have been replaced by `spacing` in the Default branding context. This PR changes the function reference and alters sizing to permitted values.

### DIFF
--- a/context/brand-context/HISTORY.md
+++ b/context/brand-context/HISTORY.md
@@ -1,5 +1,8 @@
 # History
 
+## 4.2.1 (2020-09-08)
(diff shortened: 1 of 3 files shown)


tulipdexter:
Nice 👍 . I think you're gonna need to version and add an entry in to the history as well.

sonniesedge:
@tulipdexter Thanks for that. History updated and version bumped.

tulipdexter (approved):

PR #338, opened 2020-09-08
Global status message component

matteo-gildone:


tulipdexter:
Could the readme describe what the difference between a status message that's boxed and a status message that's not boxed? For example, if you use the `--success` modifier _without_ the `--box` modifier what happens?

matteo-gildone:
added some more info about usage of the status message

tulipdexter (approved):

tulipdexter (approved):

PR #340, opened 2020-09-21
Use isOpen as source of truth

morgaan:
Use `_isOpen` state flag as the source of truth to toggle aria and class attributes as well as the trigger's label upon `init`/`open`/`close` events.


toolkits/global/packages/global-expander/js/expander.js
```diff
@@ -23,8 +24,11 @@ const Expander = class {
 		this._targetTabbableItems = makeArray(target.querySelectorAll(
 			'button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])'
 		));
-		this._isOpen = false;
-
+		if (typeof this._options.TARGET_OPEN_INITIALLY === 'boolean') {
+			this._isOpen = this._options.TARGET_OPEN_INITIALLY;
+		} else {
```
benjclark, line 29: perhaps you could remove this else statement and instead set the default value of `TARGET_OPEN_INITIALLY` to be false to save a bit of code?

toolkits/global/packages/global-expander/HISTORY.md
```diff
@@ -1,5 +1,9 @@
 # History
 
+## 2.1.0 (2020-09-21)
+    * Use isOpen as source of truth
```
benjclark, line 4: Can this elaborate a little bit more to describe what it is being used as the source of truth of?

benjclark:
I'm wondering if this option is needed. Because whether the "target" is open or closed initially could simply be down to whether or not you have the `u-js-hide` classname on it when the page loads? Same applies to the aria attributes.

toolkits/global/packages/global-expander/HISTORY.md
morgaan: Sure will do

toolkits/global/packages/global-expander/js/expander.js
morgaan: Good point 😅!

morgaan:
> I'm wondering if this option is needed. Because whether the "target" is open or closed initially could simply be down to whether or not you have the `u-js-hide` classname on it when the page loads? Same kind of thinking applies to the aria attributes.

This is expecting the component's consumer to provide 4 attributes (1 class + 3 aria) and in their appropriate initial state, where they can either rely on defaults or one single option to rule the 4 attributes in their appropriate state.

This approach of having the 4 updates to rely on 1 single state flag (internal to the component) sounds more robust than having to decipher from 4 attributes of the markup (that can have been messed up).

Furthermore by instantiating the component with the explicit option you communicate more clearly your intent I believe.

What do you think?

tulipdexter (approved):
🚀 Nice one @morgaan and @nasiamakryg !

benjclark:
For anyone reading this later that wants to know what happened with the discussion. We all had a chat about it on Slack and agreed to introduce the option when we have something that needs to use it. But this PR should still make changes to set aria attributes during the module's init.

benjclark (approved):
👍 :

PR #346, opened 2020-10-01
remove styling for statuses in c-box

ricky-lau:
global-status-message has taken over. removing statuses from box to avoid confusion.


benjclark (approved):
🧹  nice

PR #347, opened 2020-10-05
Autocomplete demo

alexkilgour:
Update demo code for `global-autocomplete` so that it (hopefully) publishes a static self-contained demo file at `demo/dist/index.html` using the latest version of the `frontend-package-manager`

* Update `global-autocomplete` to use ESM exports
* Update `global-autocomplete` demo folder code to produce a working demo
* Update to the latest version of `frontend-package-manager` for code demo compilation (and document)

Example output when generating demo locally:
Screenshot 2020-10-05 at 13 24 45


jpw (approved):
Wonderful!

PR #349, opened 2020-10-06
Add extra modifiers to springer-box

ricky-lau:
Added these two classes to support the removal of [c-panel](https://github.com/springernature/oscar/tree/master/lib/frontend/src/shared/refactor-pandora/scss/panel) in refactor-pandora folder

### DIFF
--- a/toolkits/springer/packages/springer-box/scss/50-components/_box.scss
+++ b/toolkits/springer/packages/springer-box/scss/50-components/_box.scss
@@ -10,6 +10,10 @@
 	padding: $box--padding;
 }
 
+.c-box--background-grey {
+	background-color: $box--background-color-grey;
+}
+
 .c-box__heading {
 	@include font-size($box--heading-font-size);
 	border-bottom: $box--border;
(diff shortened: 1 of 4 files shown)


toolkits/springer/packages/springer-box/scss/50-components/_box.scss
```diff
@@ -2,14 +2,20 @@
...
+	border: $box--border;
 	line-height: $box--line-height;
 	padding: $box--padding;
 }
 
+.c-box--background-grey {
+	background-color: $box--background-color-grey;
+}
+
+.c-box--borderless {
```
tulipdexter, line 14: I wonder if this is backwards and the modifiers should be `c-box--bordered` and `c-box-shadowed`

ricky-lau: I did have that thought too as it would be neater with less to override but then would it also be weird to have `c-box` not looking like a box..?

tulipdexter: Oh yeah, if the background is white then it's just nothing.
Makes sense. Although I'm not sure border and shadow are the same thing. Perhaps `c-box--shadow` could be split out?

tulipdexter (approved):

PR #352, opened 2020-10-12
Add nature footer as a component

nickcall:
First step of moving the nature footer from the nature-150 folder into a shared package.


alexkilgour:
Is the idea that the code in the README can just be dropped into any product to get the footer (if you include the CSS)? Or does it need tweaking per product

nickcall:
You will be a need to tweak the detail (link paths/text etc) but the overall structure should stay the same.

alexkilgour (approved):

PR #353, opened 2020-10-12
Adds arrow mixin

jon-stevens:
Adds a nice arrow mixing for good arrow times
Screenshot 2020-10-12 at 16 55 11
Screenshot 2020-10-12 at 16 55 07

### DIFF
--- a/context/brand-context/default/scss/30-mixins/arrow.scss
+++ b/context/brand-context/default/scss/30-mixins/arrow.scss
@@ -1,6 +1,6 @@
+@use '../20-functions/url-encode';
 
 // Variable arrow, useful for things like accordions
-
 @mixin arrow($colorHex, $direction, $size) {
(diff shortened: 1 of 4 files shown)


context/brand-context/default/scss/30-mixins/arrow.scss
```diff
@@ -0,0 +1,83 @@
...
+
+	&.position-right {
+		&:before {
+			float: right;
+			opacity: 0.5;
+		}
+	}
+}
+
+@function str-replace($string, $search, $replace: '') {
```
alexkilgour, line 44: maybe the `url-encode`/`str-replace` can go in a separate `url-encode` file that gets used here (just import it at the top of this file)? feels like that could be useful in other places

alexkilgour (approved):
👍

PR #355, opened 2020-10-14
Arrow fix

jon-stevens:
Remove css rule, and change @use to @import (doesn't seem to work - may be the version of sasss)


saramorueta (approved):
it's just a pity but... 🤷

PR #356, opened 2020-10-14
adds loader animation

jon-stevens:
Show the loader swirly thing to distract the user from the inevitable void while requests and things are happening.

loader-animation

### DIFF
--- a/toolkits/global/packages/global-loader-animation/scss/50-components/_loader.scss
+++ b/toolkits/global/packages/global-loader-animation/scss/50-components/_loader.scss
@@ -1,0 +1,71 @@
+@keyframes lds-ripple {
+	0% {
+		top: 84px;
+		left: 84px;
+		width: 0;
+		height: 0;
+		opacity: 1;
+	}
+
+	100% {
+		top: 12px;
+		left: 12px;
+		width: 144px;
+		height: 144px;
+		opacity: 0;
+	}
+}
+
+@-webkit-keyframes lds-ripple {
+	0% {
+		top: 84px;
+		left: 84px;
+		width: 0;
+		height: 0;
+		opacity: 1;
+	}
+
+	100% {
+
+		top: 12px;
+		left: 12px;
+		width: 144px;
+		height: 144px;
+		opacity: 0;
+	}
+}
+
+.lds-ripple {
+	position: fixed;
+	width: 100px;
+	height: 100px;
+	left: 50%;
+	transform: translateX(-50%);
+	-webkit-transform: translate(-100%, -100%) scale(0.5) translate(50px, 50px);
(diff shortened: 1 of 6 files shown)


toolkits/global/packages/global-loader-animation/scss/50-components/_loader.scss
```diff
@@ -0,0 +1,71 @@
...
+
+		top: 12px;
+		left: 12px;
+		width: 144px;
+		height: 144px;
+		opacity: 0;
+	}
+}
+
+.lds-ripple {
```
alexkilgour, line 38: component class needs to have a `c-` prefix

toolkits/global/packages/global-loader-animation/scss/50-components/_loader.scss
```diff
@@ -0,0 +1,71 @@
...
+	position: fixed;
+	width: 100px;
+	height: 100px;
+	left: 50%;
+	transform: translateX(-50%);
+	-webkit-transform: translate(-100%, -100%) scale(0.5) translate(50px, 50px);
+	transform: translate(-100%, -100%) scale(0.5) translate(50px, 50px);
+	z-index: $zindex-level-highest;
+
+	div {
```
alexkilgour, line 48: As this is global, we should have a `10-settings/_default.scss` file that sets some standard settings for this, then if this is SN a `10-settings/_springernature.scss` file that brands those setting for SN.

Candidates for abstracting to settings would be `border-width` and the two instances of `border-color`.

See an example of how it's done [in this component](https://github.com/springernature/frontend-toolkits/tree/master/toolkits/global/packages/global-card/scss/10-settings) with the brand settings importing default.

toolkits/global/packages/global-loader-animation/scss/50-components/_loader.scss
```diff
@@ -0,0 +1,73 @@
+@import '../10-settings/default';
```
alexkilgour, line 1: You don't need to do this, they get imported separately in the app. where you have `// Include this with your other components` in the README, just include a line before the component import that imports the settings file

toolkits/global/packages/global-loader-animation/scss/50-components/_loader.scss
```diff
@@ -0,0 +1,73 @@
...
+		animation: lds-ripple 1.9s cubic-bezier(0, 0.2, 0.8, 1) infinite;
+		width: $size;
+		height: $size;
+
+		&:nth-child(1) {
+			border-color: $main-color;
+		}
+
+		&:nth-child(2) {
+			border-color: $secondaey-color;
```
alexkilgour, line 67: `secondary` not `secondaey`

alexkilgour (approved):
Looks good, thanks @jon-stevens 🙇‍♂️

PR #357, opened 2020-10-15
Update user details

jon-stevens:
Remove some SCSS deps, as they will be imported in the consuming app


alexkilgour:
The only issue I can see is around utility classes. We try not to use them in components, preferring to use mixins and settings level config.

You have some utility level classes inside the component - `u-ml10`, `u-font-12`, `u-color-darker-grey`, `u-mb10` - that could be incorporated into existing classes and then configured via variables at the `10-settings` level.

You then use `layout` and `positioning` from the brand-context - however utilities are the only level that are **not**  included with the context. The idea with utilities is that they override component classes, so are included in your app at the very end. Any utility that is more than a single line declaration should have an accompanying mixin as in [this example](https://github.com/springernature/frontend-toolkits/blob/master/context/brand-context/default/scss/60-utilities/container.scss). The mixins come with context so you can use them in your component that way. For example, where you have used `brand-context/default/scss/30-mixins/arrow/scss` this is correct, as it comes with the `brand-context`.

In summary, it is preferable to refactor out the use of utilities in components, and remove them from the `view` layer. Apologies, you weren't to know this, it will be in the documentation!

alexkilgour (approved):
👌

PR #359, opened 2020-10-16
move a couple of style to core, add class

jon-stevens:
Adding back a class that I shouldn't have removed.
Moving a couple of css rules over to core.scss from enhanced.scss.


toolkits/springernature/packages/springernature-user-details/view/user-details.hbs
```diff
@@ -1,5 +1,5 @@
 <section class="c-user-details">
-    <button data-component-user-details-open class="c-user-details__open" data-expander data-expander-target="#user">
+    <button data-component-user-details-open class="c-user-details__open position-right" data-expander data-expander-target="#user">
```
alexkilgour, line 2: where is `position-right` defined?

jon-stevens: That is used by the arrow mixin. It's not the best name really, I could do with changing that

alexkilgour (approved):

PR #360, opened 2020-10-19
Changes Global Expander's FOCUS_EVENT option to OPEN_EVENT

benjclark:
At the moment Oscar repo is the only consumer of Global Expander that uses the FOCUS_EVENT option. We have a bug on Oscar at the moment which has led me to the realisation that we do not need Global Expander to focus on an element once it has done its thing. But there is still a requirement to know when Global Expander has finished opening a toggle-able element, therefore an event dispatched by Global Expander is still needed.

Therefore this PR changes the event to be what Oscar needs. Plus, it is likely to be a more useful event to any other consumer of the component. As the option now means that it is sending an event when it opens a toggle-able element, rather than an event specifically concerned with when it focused on a target element during the open process.

Relates to PR https://github.com/springernature/frontend-toolkits/pull/361

### DIFF
--- a/toolkits/global/packages/global-expander/HISTORY.md
+++ b/toolkits/global/packages/global-expander/HISTORY.md
@@ -1,7 +1,7 @@
 # History
 
 ## 3.0.0 (2020-10-19)
-    * Changes FOCUS_EVENT to OPEN_EVENT as open event is what is needed on Oscar components at this time and for any future consumer this will likely be a more useful event than the focus event in any case.
+    * Changes FOCUS_EVENT to OPEN_EVENT.
 
 ## 2.1.0 (2020-09-21)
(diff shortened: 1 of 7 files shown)


toolkits/global/packages/global-expander/HISTORY.md
```diff
@@ -1,5 +1,8 @@
 # History
 
+## 3.0.0 (2020-10-19)
+    * Changes FOCUS_EVENT to OPEN_EVENT as open event is what is needed on Oscar components at this time and for any future consumer this will likely be a more useful event than the focus event in any case.
```
tulipdexter, line 4: I don't think you need to mention Oscar here

tulipdexter (approved):

PR #369, opened 2020-11-04
Remove use of rfs font-size mixin in global components

ricky-lau:
Replace font-size with existing static value in settings


benjclark (approved):
Nice. Does this relate to your Trello card? If so can you update the outcome in there? https://trello.com/c/5Mfbn6Tv/66-any-pushback-on-adding-responsive-font-size-mixin-into-nature
👍

ricky-lau:
indeed it does and thanks for reminding about the trello card

PR #376, opened 2020-11-12
update nature settings for global-meta

ricky-lau:
Update Nature settings to match existing c-meta style from app (and support it being removed from nature app)

### DIFF
--- a/toolkits/global/packages/global-meta/scss/10-settings/_nature.scss
+++ b/toolkits/global/packages/global-meta/scss/10-settings/_nature.scss
@@ -1,14 +1,17 @@
 /**
  * Meta
- * Springer branding for component
+ * Nature branding for component
  */
 
 // Default button setting
 @import 'default';
 
-$meta--font-size: 1.3rem;
+$meta--font-size: 1.2rem;
 $meta--font-family: $font-family-sans;
-$meta--border: 1px solid greyscale(80);
-$meta--color: greyscale(40);
-$meta--margin-bottom: 4px;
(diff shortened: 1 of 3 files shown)


toolkits/global/packages/global-meta/scss/10-settings/_nature.scss
```diff
@@ -1,14 +1,17 @@
...
 $meta--font-family: $font-family-sans;
-$meta--border: 1px solid greyscale(80);
-$meta--color: greyscale(40);
-$meta--margin-bottom: 4px;
+$meta--border: 1px solid #bbb;
+$meta--color: inherit;
+$meta--line-height: 1.4;
+$meta--margin-bottom: spacing(4);
+$meta--margin-right: spacing(0);
+$meta--padding-right: spacing(4);
```
tulipdexter, line 16: So Nature meta will _always_ want this padding to the right?

ricky-lau: Yeah and the component wont apply this to the last child.

Unless you mean we have a situation ive missed where the meta is right aligned?

tulipdexter: This is spacing between meta things? Why padding and not margin?

ricky-lau: Oh I see. It's for the spacing of border that separates between items

tulipdexter (approved):

PR #387, opened 2020-11-16
Brand Context: Update Springer typography to use `u-` prefix on headings

tulipdexter:


ricky-lau (approved):

context/brand-context/springer/scss/40-base/typography.scss
```diff
@@ -12,24 +12,24 @@ h4,
 
 h1,
 .h1 {
```
alexkilgour, line 14: should these classes not be removed here?
e.g `.h1` not needed as we now have `.u-h1`

alexkilgour (approved):

PR #390, opened 2020-11-19
Package Batch Patch

tulipdexter:
Bumps packages to bring in the [latest changes to the package-manager postscript](https://github.com/springernature/frontend-package-manager/pull/58)  which fixes a problem with `npx`.

Some packages are patch bumps just to bring in those changes.
Others are bumps to old contexts but will still bring in the latest package manager change.


benjclark (approved):
oh my gawd this is so much fun

PR #397, opened 2020-11-26
Updates documentation

anibe:
The code example was missing `springernature-` from the filenames


alexkilgour:
Probably worth updating the patch version so that this change is published. The README is used by the styleguide to generate documentation (which it gets via NPM), so without pushing a new version the styleguide docs will be out of date

anibe:
@alexkilgour My bad. Will update

alexkilgour:
> @alexkilgour My bad. Will update

It's all good, thanks 😃

alexkilgour (approved):

PR #402, opened 2020-12-01
[JaeRae/AW] handle displaying leaderboard conditionally based on inst…

graearea:
…itutional banner

### DIFF
--- a/toolkits/global/packages/global-ad/HISTORY.md
+++ b/toolkits/global/packages/global-ad/HISTORY.md
@@ -1,5 +1,8 @@
 # History
 
+## 1.1.0 (2020-12-01)
+	* Handle conditionally displaying leaderboard ad when there is no institutional access banner
+
 ## 1.0.1 (2020-11-18)
     * Bump to get latest version package-manager with updated post install script
 
--- a/toolkits/global/packages/global-ad/package.json
+++ b/toolkits/global/packages/global-ad/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@springernature/global-ad",
-  "version": "1.0.1",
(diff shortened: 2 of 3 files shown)


benjclark:
Does this mean that, for the ads we do not want to display, we are still sending http requests for the ad and rendering the ad in to the ad slot but we then hide it with css?

moddular:
> Does this mean that, for the ads we do not want to display, we are still sending http requests for the ad and rendering the ad in to the ad slot but we then hide it with css?

It shouldn't do I think, we don't (or shouldn't) trigger loading the ad if it isn't visible. So the idea is the leaderboard is display: none by default - and doesn't load anything. If there is no institutional banner the component will return an element with the u-show-following-ad class - which causes the ad to be displayed and triggers the loading.

benjclark:
> > Does this mean that, for the ads we do not want to display, we are still sending http requests for the ad and rendering the ad in to the ad slot but we then hide it with css?
> 
> It shouldn't do I think, we don't (or shouldn't) trigger loading the ad if it isn't visible. So the idea is the leaderboard is display: none by default - and doesn't load anything. If there is no institutional banner the component will return an element with the u-show-following-ad class - which causes the ad to be displayed and triggers the loading.

Ok great. Yeah, I was wondering if we needed to include some logic in the GPT script in GTM to say don't load this ad slot if classname x exists on the parent element or something like that. But if GPT doesn't load ads if they are display: none by default then all good!

benjclark (approved):

moddular:
I believe it's true - e.g. I don't think we load anything on mobiles when the ad is hidden. If it's not true, then I'll fix the GPT script :)

PR #412, opened 2020-12-15
remove rfs mixin from springer packages

ricky-lau:
Replace font-size with existing static value in settings

https://trello.com/c/5Mfbn6Tv


benjclark (approved):
Goodbye, dear friend

PR #414, opened 2020-12-16
Adds modifier for keyline on first list item only

benjclark:
We have lists where items can appear on a list depending on whether they have been enabled in Pubserv or not. This means it's impossible to know which item to put a keyline on. This classname allows us to apply to keyline to the first list item only not matter what config has been saved in Pubserv for a journal


ricky-lau:
this might be helpful to have in the readme 😁

benjclark:
omg true

tulipdexter (approved):

PR #415, opened 2020-12-17
Adds important to hide utilities

benjclark:
Had an issue today where our u-hide utility classnames were not working because of selector specificity, i.e. we had some css that set display: block and because of the specificity of it it trumped the hide utility. I think it makes sense to add important to hide utilities in general. If we've made the conscious choice to add a classname like u-hide to the html then it should be a certainty that the element is hidden.

### DIFF
--- a/context/brand-context/HISTORY.md
+++ b/context/brand-context/HISTORY.md
@@ -1,5 +1,8 @@
 # History
 
+## 6.3.0 (2020-12-11)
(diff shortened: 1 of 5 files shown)


alexkilgour (approved):
I do think there is actually a case for making _all_ utilities use `!important`. When discussing this originally, it was felt we should need to - because of the layered way the CSS works, with utilities being last, if it is followed strictly then there should be no need. But I think the reality is different. I know `!important` is seen as _bad_, but I do think there is a very simple justification for it here

PR #417, opened 2021-01-04
Move font-weight utilities to default context

morgaan:
... and remove from Springer brand

### DIFF
--- a/context/brand-context/HISTORY.md
+++ b/context/brand-context/HISTORY.md
@@ -1,7 +1,7 @@
 # History
 
 ## 7.0.0 (2021-01-04)
-    * BREAKING: Move font-weight utilities to default context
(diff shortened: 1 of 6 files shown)


context/brand-context/HISTORY.md
```diff
@@ -1,5 +1,8 @@
 # History
 
+## 7.0.0 (2021-01-04)
+    * BREAKING: Move font-weight utilities to default context
```
alexkilgour, line 4: Can we mention here that it is moving out of springer, something like
`BREAKING: font-weight utilities moved from SPRINGER to DEFAULT context`
Will help with scanning

morgaan: Most certainly 👍

alexkilgour (approved):
Great 👌

PR #419, opened 2021-01-13
Tidy up nature-header at narrow screen widths

ricky-lau:
https://jira.springernature.com/browse/PAN-1068

### Before
www nature com_(iPhone 5_SE)

### After
local-www nature com_7890_nature(iPhone 5_SE) (3)

### DIFF
--- a/toolkits/nature/packages/nature-header/HISTORY.md
+++ b/toolkits/nature/packages/nature-header/HISTORY.md
@@ -5,7 +5,7 @@
     * Updated README with updated example
     * Add back keyline for expander (used in search dropdown)
 
-## 3.2.0 (2020-01-07)
+## 3.2.0 (2021-01-07)
     * Add font weight to c-header__menu--secondary 
 
 ## 3.1.0 (2020-12-16)
--- a/toolkits/nature/packages/nature-header/HISTORY.md
+++ b/toolkits/nature/packages/nature-header/HISTORY.md
@@ -1,5 +1,10 @@
 # History
 
+## 4.0.0 (2020-01-13)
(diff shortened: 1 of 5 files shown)


toolkits/nature/packages/nature-header/scss/50-components/header.scss
```diff
@@ -53,11 +52,25 @@
...
+		}
+	}
+}
+
+.c-header__menu--journal {
 	font-weight: $font-weight-bold;
 }
 
-.c-header__menu--secondary {
+.c-header__menu--updates {
```
tulipdexter, line 73: What does 'updates' mean in the header?

ricky-lau: it was an attempt to group the sign up for alerts and rss feed links together

tulipdexter: I was trying to group that mean and the journal menu as the same 'thing'. That's why I went with secondary. Stylistically they're the same.

tulipdexter: But if you're keen to split them, maybe something like `--tools` (as we moved them from journal tools section

toolkits/nature/packages/nature-header/HISTORY.md
```diff
@@ -1,5 +1,10 @@
 # History
 
+## 4.0.0 (2020-01-13)
```
tulipdexter, line 3: 2021 🎉 🍾 🎆

tulipdexter (approved):

PR #420, opened 2021-01-14
Removes important from hiding utils as not needed any longer

benjclark:


alexkilgour:
What's the reasoning for not needing this anymore?

benjclark:
@alexkilgour discussed with colleagues and we'd prefer to refactor existing css to not require specificity to work. Also, the issue that this solved is no longer an issue if this !important is removed

jpw (approved):

PR #425, opened 2021-01-21
dropdown is full width on narrow viewport

ricky-lau:
* Expander full width at narrow viewports
* Aligned last expander in list of dropdown to right of screen to prevent horizontal scroll
* Reduced font-size and space between chevrons on journal menu items to get all on one line for 320px/'xs' breakpoint and above    

### **Before**
www nature com_srep_(iPhone 5_SE)

### **After**
local-www nature com_7890_srep(iPhone 5_SE)


tulipdexter (approved):
All seems wonderfully sensible to me 🚀

PR #429, opened 2021-01-26
Move springer-lists component to global

nasiamakryg:
- Add springernature theme
 - Rename mixin `interface-link` to `u-link-interface`and move it to global brand-context

### DIFF
--- a/toolkits/global/packages/global-lists/package.json
+++ b/toolkits/global/packages/global-lists/package.json
@@ -1,0 +1,9 @@
+{
+  "name": "@springernature/global-lists",
+  "version": "1.0.0",
+  "license": "MIT",
+  "description": "Provide different styled lists",
+  "keywords": [],
+  "author": "Springer Nature",
+  "brandContext": "^6.0.0"
+}
--- a/toolkits/global/packages/global-lists/scss/10-settings/_default.scss
+++ b/toolkits/global/packages/global-lists/scss/10-settings/_default.scss
@@ -1,0 +1,17 @@
+$global-list-columned--column-gap: spacing(24);
+$global-list-columned--margin-bottom: 0.25em;
+
+$global-list-description--wrapper-margin-bottom: spacing(24);
+$global-list-description--font-weight: $font-weight-bold;
+$global-list-description--font-size: $font-size-sm;
+$global-list-description--line-height: 1.4;
+$global-list-description--margin-bottom: 0.25em;
+$global-list-description--is-interface-link: true;
+
+$global-list-group--border-color: greyscale(80);
+$global-list-group--padding-xs: spacing(4) 0;
+$global-list-group--padding-sm: spacing(8) 0;
+$global-list-group--padding-md: spacing(16) 0;
+$global-list-group--padding-lg: spacing(24) 0;
+$global-list-group--line-height: 1.4;
+$global-list-group--is-interface-link: true;
--- a/toolkits/global/packages/global-lists/scss/10-settings/_default.scss
+++ b/toolkits/global/packages/global-lists/scss/10-settings/_default.scss
@@ -1,17 +1,17 @@
-$global-list-columned--column-gap: spacing(24);
+$global-list-columned--column-gap: 24px;
 $global-list-columned--margin-bottom: 0.25em;
 
-$global-list-description--wrapper-margin-bottom: spacing(24);
+$global-list-description--wrapper-margin-bottom: 24px;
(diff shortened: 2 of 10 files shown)


toolkits/global/packages/global-lists/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "@springernature/global-lists",
+  "version": "1.0.0",
+  "license": "MIT",
+  "description": "Provide different styled lists",
+  "keywords": [],
+  "author": "Springer Nature",
+  "brandContext": "^6.0.0"
```
alexkilgour, line 8: Current version is `8.0.1`, is there a reason for using `6.0.0`?

toolkits/global/packages/global-lists/scss/50-components/_list-description.scss
```diff
@@ -0,0 +1,39 @@
+.c-list-description {
+	font-size: $global-list-description--font-size;
+	font-family: $font-family-sans;
+	line-height: $global-list-description--line-height;
+	margin-bottom: 0;
+	width: 100%; // fix text breaking layout in IE10
+
+	@if $global-list-description--is-interface-link {
```
alexkilgour, line 8: 👌 This is going to be `true` for the springernature brand as well, because it will inherit from default, is that desired behaviour?

toolkits/global/packages/global-lists/scss/10-settings/_default.scss
```diff
@@ -0,0 +1,17 @@
...
+$global-list-description--margin-bottom: 0.25em;
+$global-list-description--is-interface-link: true;
+
+$global-list-group--border-color: greyscale(80);
+$global-list-group--padding-xs: spacing(4) 0;
+$global-list-group--padding-sm: spacing(8) 0;
+$global-list-group--padding-md: spacing(16) 0;
+$global-list-group--padding-lg: spacing(24) 0;
+$global-list-group--line-height: 1.4;
+$global-list-group--is-interface-link: true;
```
alexkilgour, line 17: Are these the styles from the springer brand? We need to have `10-settings/_springer.scss` as well for them to migrate to, otherwise if that brand needs to change then they will need to change default which will affect all the other brands.

Make the `default` really generic, don't use mixins but use explicit values e.g. `$global-list-group--border-color: #eee;`

toolkits/global/packages/global-lists/package.json
morgaan: Maybe copy paste from `springer-lists`, I can fix that

toolkits/global/packages/global-lists/scss/50-components/_list-description.scss
morgaan: Well so far we have no nested link AFAIK, let me ask James, in case this would happen what he would prefer

toolkits/global/packages/global-lists/scss/10-settings/_default.scss
morgaan: Gotcha, will do thank you

toolkits/global/packages/global-lists/scss/10-settings/_springernature.scss
```diff
@@ -1,3 +1,12 @@
+/**
+ * List
+ * Springer branding for component
```
alexkilgour, line 3: should say "Springer Nature branding" here

toolkits/global/packages/global-lists/package.json
alexkilgour: Thinking about this again, I think we should leave this, so that Springer has a version tied to the brand that they are using. Then can release another version using the latest

toolkits/global/packages/global-lists/scss/10-settings/_springernature.scss
morgaan: God damn it bloody copy paste, will fix it

alexkilgour (approved):
Looking good, thanks both 🙏

toolkits/global/packages/global-lists/package.json
tulipdexter: It depends if you're deleting the springer-lists package. If not, then our usage of that will keep us tied to that 6.0.0 brand context.

tulipdexter: Then we can check any issues with swapping to the global one and delete the springer-lists if all is good?

PR #433, opened 2021-01-29
Remove RFS from brand-content

ricky-lau:
Reverts mostly https://github.com/springernature/frontend-toolkits/pull/297/files

- havent bumped version yet to coordinate releasing along with another breaking change


context/package-manager.json
```diff
@@ -21,11 +21,6 @@
       "svg",
       "ttf",
       "json"
-    ],
```
alexkilgour, line 24: Can we leave this in place as this is just saying that a test folder is allowed, which is still true

package.json
```diff
@@ -49,7 +49,6 @@
     "@springernature/frontend-package-manager": "^6.1.3",
     "@springernature/sasslint-config": "^1.2.1",
     "@springernature/util-eslint-runner": "^0.0.3",
-    "@springernature/util-sass-compiler": "^0.1.0",
```
alexkilgour, line 52: Can we also leave this and I am actually going to be using it in an upcoming PR!

alexkilgour (approved):

PR #435, opened 2021-02-01
context: update vars to consistently use context-- prefix

tulipdexter:


tulipdexter:
This PR replaces a large number of variable names from the brand context, and as such will cause breakages in many components.
These components will shortly receive a major release to fix any references to the old variable names.

alexkilgour (approved):
Nice work ✨

PR #438, opened 2021-02-01
Hotfix global lists sass

nasiamakryg:
- Fix the if `is-interface` condition by adding comparison operator
- Pull `u-text-interface` out of the condition

### DIFF
--- a/toolkits/global/packages/global-lists/scss/50-components/_list-description.scss
+++ b/toolkits/global/packages/global-lists/scss/50-components/_list-description.scss
@@ -1,7 +1,7 @@
 .c-list-description {
+	@include u-text-interface;
 	margin-bottom: 0;
 	width: 100%; // fix text breaking layout in IE10
-	@include u-text-interface;
 
 	@if ($global-list-description--is-interface == true) {
 		a {
--- a/toolkits/global/packages/global-lists/scss/50-components/_list-description.scss
+++ b/toolkits/global/packages/global-lists/scss/50-components/_list-description.scss
@@ -1,10 +1,9 @@
 .c-list-description {
 	margin-bottom: 0;
 	width: 100%; // fix text breaking layout in IE10
-	
-	@if $global-list-description--is-interface {
-		@include u-text-interface;
+	@include u-text-interface;
 
+	@if ($global-list-description--is-interface == true) {
 		a {
 			@include u-link-interface;
 		}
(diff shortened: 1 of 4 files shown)


toolkits/global/packages/global-lists/scss/50-components/_list-description.scss
```diff
@@ -1,10 +1,9 @@
 .c-list-description {
 	margin-bottom: 0;
 	width: 100%; // fix text breaking layout in IE10
-	
-	@if $global-list-description--is-interface {
-		@include u-text-interface;
+	@include u-text-interface;
```
morgaan, line 4: Hey Nasia nicely spot for the if condition!

Though, wasn't it the core issue? If yes, then I would still wrap the `u-text-interface` into a if block, though at line 2 rather than 4, in order to avoid failing the build because of "Mixins should come before declarations" (see: https://github.com/sasstools/sass-lint/blob/master/docs/rules/mixins-before-declarations.md).

morgaan: If you go for the change, think about updating the HISTORY entry ;)

morgaan: In any case the "Mixins should come before declarations" must be fixed, to fix the failing build

nasiamakryg: The thinking behind the `u-text-interface` out of the condition ha to do with the fact that we always need to have the fonts of the content of the list styled, no matter what happens with the link styling (if there is any). And according to James, they always need to be sans (for all brands actually). That's why I took it out of the condition. To give the opportunity to the text to be styled explicitly than inheriting font styling that we won't want.

As for the priority of the mixin thank you for spotting it. i will write it first ;)

alexkilgour (approved):
Good spot. I'll take the blame for this one 🙈

morgaan:
I would rather make explicit set of the content font styling... than use a mixin that was intended for something else. If someone happens to change the style of text-interface, this would break our component. Don't you think? Separation of concerns FTW IMO

toolkits/global/packages/global-lists/scss/50-components/_list-description.scss
morgaan: See answer later down the thread, sorry if I broke the thread 🤦

morgaan:
> I would rather make explicit set of the content font styling... than use a mixin that was intended for something else. If someone happens to change the style of text-interface, this would break our component. Don't you think? Separation of concerns FTW IMO

It turns out that this is inherited from springer-lists component, so let it be like this and we'll see if it becomes a problem in the future

PR #440, opened 2021-02-03
remove aria-pressed and replace aria-hidden with hidden attribute in global-expander

ricky-lau:
Issues raised from nature VPAT review:

https://jira.springernature.com/browse/PAN-1145
https://jira.springernature.com/browse/PAN-1147


hollsk (approved):
🎉

hollsk (approved):
Nice! I was totally paying attention when I last approved that!

PR #453, opened 2021-02-05
brand-context: reorder abstract imports

tulipdexter:
This ensures that the dependency settings always come first.


alexkilgour (approved):
Looking good 👍

PR #445, opened 2021-02-03
[MC, NM] WiP Add SpringerNature brand context

morgaan:
This is taken from work that has been done on Researcher Portal and approved by @foxintherain .

- Paste all context from Researcher Portal
- Try to merge latest updates made to the frontend-toolkit

PR #454, opened 2021-02-08
[NM, MC] Add new Springernature context

morgaan:
This is taken from work that has been done on Researcher Portal and approved by @foxintherain .

- Paste all context from Researcher Portal
- Try to merge latest updates made to the frontend-toolkit

This PR is the spin off of  #445 (in case you need to access the discussion thread) and concerns only changes related to Brand Context (none of components changes)


context/brand-context/default/scss/10-settings/buttons.scss
```diff
@@ -16,6 +16,7 @@ $context--button-font-size-normal: 1.6rem;
 $context--button-font-size-large: 1.8rem;
 $context--button-icon-margin: spacing(8);
 $context--button-padding: spacing(8);
+$context--button-padding-small: 0;
```
alexkilgour, line 19: Previous implementation of `@button-small` did not change the padding, so this is going to break any previous use of this. The `u-button-small` class is used in combination with the `u-button` class, so previously padding would come from the use of `u-button`. Make this value `inherit` and override (as you have) in brand settings to make this a non-breaking change

context/brand-context/springernature/scss/60-utilities/buttons.scss
```diff
@@ -0,0 +1,31 @@
...
+		box-shadow: none;
+		outline: none;
+	}
+}
+
+// Button themes
+
+.u-button--primary {
+	@include button-primary
+}
```
alexkilgour, line 22: this is already defined in `default/scss/60-utilities/buttons.scss`

context/brand-context/springernature/scss/60-utilities/buttons.scss
```diff
@@ -0,0 +1,31 @@
...
+
+.u-button--secondary {
+	@include button-secondary;
+
+	&:hover {
+		box-shadow: none;
+		border: $context--button-border-width-style $context--button-brand-color;
+	}
+}
```
alexkilgour, line 31: we already have `default/scss/60-utilities/buttons.scss`. I think we need to find a way to incorporate these extra styles into the settings so that you only need to include the utility once (from default). We want to keep utilities as much as possible in the `default` brand, and we don't want to be requiring you include two files, one from each brand as there is a high possibility of user error.

context/brand-context/springernature/scss/40-base/layout.scss
```diff
@@ -12,10 +12,11 @@ html {
...
 	font-family: $context--font-family-serif;
-	line-height: 1.76;
-	color: color('mineshaft');
-	background: color('alabaster');
-	font-size: 1.7em; /* fixes chrome rems bug - http://stackoverflow.com/questions/20099844/chrome-not-respecting-rem-font-size-on-body-tag */
+	font-size: $context--font-size-base;
+	line-height: $context--line-height-base;
+	color: $context--primary-text-color;
+	background: $context--page-background-color;
+	font-size: $context--font-size-base;
```
alexkilgour, line 19: you've defined `font-size` twice

context/brand-context/default/scss/30-mixins/buttons.scss
```diff
@@ -115,11 +117,14 @@
...
 	&:hover {
 		border: map-get($theme, 'hoverBorder');
+		box-shadow: map-get($theme, 'hoverBoxShadow');
 		text-decoration: map-get($theme, 'hoverTextDecoration');
 	}
 
 	&:focus {
 		border: map-get($theme, 'focusBorder');
+		box-shadow: map-get($theme, 'focusBoxShadow');
+		outline: map-get($theme, 'focusOuline');
```
nasiamakryg, line 127:
```suggestion
		outline: map-get($theme, 'focusOutline');
```

alexkilgour (approved):
Thank you for doing this. Really great work 😄

PR #455, opened 2021-02-08
[NM, MC] Add new Springernature components brand

morgaan:
This is taken from work that has been done on Researcher Portal and approved by @foxintherain .

- Paste all context from Researcher Portal
- Try to merge latest updates made to the frontend-toolkit

This PR is the spin off of  #445 (in case you need to access the discussion thread) and concerns only changes related to components (none of brand context changes)


alexkilgour:
Where you have added new settings for `springernature` can you make sure that you update the README to detail this in the branding section, for example here: https://github.com/springernature/frontend-toolkits/tree/master/toolkits/global/packages/global-breadcrumbs#branding

toolkits/springernature/packages/springernature-header/scss/50-components/core.scss
```diff
@@ -33,19 +33,19 @@
...
 
-	@include media-query('tablet') {
+	@include media-query('desktop-width') {
 		font-size: 2.7rem;
 	}
 
 	font-family: $context--font-family-serif;
 	line-height: 4.6rem;
 	font-size: 2rem;
 	font-size: 1.5rem;
```
alexkilgour, line 47: You have defined `font-size` twice here. Also, this is the `core` endpoint which means it needs to support _all_ browsers, we should not be using REMs for font-size. These should be pixel values that get updated to REM's in the `enhanced` endpoint

toolkits/springernature/packages/springernature-header/scss/50-components/core.scss
```diff
@@ -33,19 +33,19 @@
...
-	color: color('mineshaft');
+	color: color('black');
 
 	&.hide {
```
alexkilgour, line 50: this is the wrong naming convention, should be a modifier `c-header__journal-title--hide`

toolkits/springernature/packages/springernature-user-details/package.json
```diff
@@ -1,11 +1,11 @@
...
-  "version": "1.3.4",
+  "version": "2.0.0",
   "license": "MIT",
   "description": "Springer Nature branded logged-in user details component",
   "keywords": [],
   "author": "Springer Nature",
-  "brandContext": "^4.3.1",
+  "brandContext": "^10.0.0",
   "dependencies": {
     "@springernature/global-expander": "^2.0.2"
```
alexkilgour, line 10: we shouldn't be doing this. Components should not depend on other components. But leave it here for now and I will think how to tackle this.

toolkits/springernature/packages/springernature-header/scss/50-components/core.scss
morgaan: It was there already but I can fix it sure

toolkits/springernature/packages/springernature-header/scss/50-components/core.scss
morgaan: It was there already but I can fix it sure. At least try

toolkits/springernature/packages/springernature-user-details/package.json
morgaan: It was there already and it intrigued us. Glad you spotted it. So I believe I resolve the conversation on that one and I assume you have it on your todo somewhere

toolkits/springernature/packages/springernature-header/scss/50-components/core.scss
morgaan: Ok TBH we did not authored this package nor its latest changes. We are just patching colour and breakpoint keys in this PR. Hence, so far we do not use this package on the RP project (this will come).
Since I do not want to make false assumptions of what the pixel values... should be, I would rather leave @sky-jack taking care of this in a separate PR. What do you think @alexkilgour ?

toolkits/springernature/packages/springernature-header/scss/50-components/core.scss
morgaan: Same as above but in regards to the `.hide` class to become `.c-header__journal-title--hide`. Maybe a job for @sky-jack in another PR. What do you think @alexkilgour ?

toolkits/springernature/packages/springernature-header/scss/50-components/core.scss
alexkilgour: OK, that makes sense. Fine to leave that here. Leave it with me and I will talk to jack on Thursday

toolkits/springernature/packages/springernature-header/scss/50-components/core.scss
alexkilgour: I actually think in this case we should update the class. It's a simple change, and we are already making a breaking change to the component. If we don't then it's just another breaking change straight afterwards. Just change the class and detail in the HISTORY file that `hide` has changed to `xyz`

alexkilgour (approved):
Looking good 🌈

PR #456, opened 2021-02-11
[NM, MC] Fix button box-shadow on hover for Springernature brand

nasiamakryg:
Delete inherit from box-shadow property on hover since it doesn't really inherit anything 🙈


morgaan (approved):
Great job spotting that! Did bump the version after merging back master in ;)

PR #459, opened 2021-02-12
[NM, MC] Re-introduce global card title font-weight settings

morgaan:
* BREAKING:
	* Switch to use new `brand-context` dependency
	* Remove springer settings as it does not override anything any more
* FEATURE:
	* Re-introduce the variable `$card--title-font-weight`
	* Add branding configurations for springer nature


toolkits/global/packages/global-card/scss/10-settings/springer.scss
```diff
@@ -1,10 +0,0 @@
-/**
- * Card
- * Springer branding for component
- */
-
-// Default setting
-@import 'default';
-
-// Springer overrides
-$card--title-font-weight: normal;
```
nasiamakryg, line 10: doesn't that setting need to be preserved?

toolkits/global/packages/global-card/scss/50-components/card.scss
```diff
@@ -33,6 +33,9 @@
 
 .c-card__title {
 	@include u-h3;
+	@if $card--title-font-weight != null {
+		font-weight: $card--title-font-weight;
+	}
```
alexkilgour, line 38: Is there a reason that this needs to be set to null by default? Could we not make `$card--title-font-weight` be `inherit` then we don't need this conditional?

toolkits/global/packages/global-card/scss/10-settings/springer.scss
morgaan: Actually not according to @tulipdexter

toolkits/global/packages/global-card/scss/50-components/card.scss
morgaan: Well since the mixin sets `font-weight` already, this would result in such CSS output:

```css
  ...
  font-weight: bold;
  font-weight: inherit;
  margin-bottom: $card--spacing;
```

Where with the `null` we do not duplicate the prop unless we have a reason to

alexkilgour: so it looks like it's the `u-h3` mixin, but that doesn't exist at the default level so this is going to break anyway

tulipdexter: Ah that's interesting. Those mixins should probably go into the default context tbh.

nasiamakryg: I must admit that I am not a fan of the u-h* mixins being used in cases as this one. 
One of the reasons is that each brand defines the heading styles differently (especially the font-weight for them) and they aslo define and use "cards" differently. The fact that the cards use headings as titles doesn't mean that they wanna have the headings style as well. I suppose that's why you see in both springer and nature the c-card__title being used on an element together with other utility classes which override properties of the first. 
I understand the effort to maintain consistency but I feel we are not ready yet to dictate styles in this level.

morgaan: @alexkilgour I bumped the brand context to be using version from your latest PR. So I believe we are good to merge AFAIK?

tulipdexter: Yeah the decision to use `u-h3` here was an attempt to have a single source of truth for headings, which is defined in the typography. I was concerned that each component could define its own title / heading styles and consistency could easily drift.
I would be interested in the idea of UXD coming up with a set of _other_ heading styles that could be defined in the typography. That would leave the `u-h*` to be considered for 'copy' type things, and maybe something like `@include interface-heading` for other headings? It could take a param of the intensity, `@include interface-heading(1)`.

alexkilgour (approved):
Looking good now

PR #460, opened 2021-02-12
global-expander bugs

ricky-lau:
* BUG: 
    * Fix expander incorrectly closing when shift+tab on last item and there are still tabbable items before it
    * Fix expander not closing when tababble number of items in expander changes
    * Bump global-javascript to 3.0.1


benjclark (approved):
Nice!

PR #461, opened 2021-02-12
Update default headings to use mixins

alexkilgour:
All brands except the `default` brand were using a mixin for the heading utility classes e.g `u-h1`. This causes an issue for components that want to reference that mixin.

FIX: all the mixin to the `default` brand level, to be overridden by each imprint brand


morgaan (approved):
Thank you Alex 🏅

nasiamakryg (approved):
https://media.giphy.com/media/Hw8vYF4DNRCKY/giphy.gif

PR #462, opened 2021-02-15
Fix nature-hero image not appearing in IE10

ricky-lau:
bugbash item: https://trello.com/c/ZjjBiBq4/46-ie10-hero-image-does-not-appear-in-ie10


benjclark (approved):
Thanks Ricky :-)

PR #467, opened 2021-02-17
Button mixin svg fill

nasiamakryg:
This change will make sure that any svg icon included on a button gets the currentColor used by the font.

Screen Shot 2021-02-17 at 14 40 13


alexkilgour (approved):
Good spot 😄

PR #468, opened 2021-02-17
Update global popup to allow for wrapped text

nickcall:
- Positions the popup and arrow to be over the trigger when the trigger is positioned at the "end of line" and wraps.
- Ensures that the popup doesn't go off the screen.
- Ensures the popup appears at the correct size and isn't shrunk or stretched.
- New version number added

### DIFF
--- a/toolkits/global/packages/global-popup/js/popup.js
+++ b/toolkits/global/packages/global-popup/js/popup.js
@@ -42,7 +42,7 @@ const Popup = class {
 		const pos = this._calcPositioning();
 		this._content.style.top = this._px(pos.top);
 		this._content.style.left = this._px(pos.left);
-		this._content.style.right = this._px(5);
+		this._content.style.right = this._px(pos.right);
 	}
 
 	_close() {
@@ -96,20 +96,21 @@ const Popup = class {
 
 	_calcPositioning() {
 		const distanceScrolled = document.documentElement.scrollTop;
+		const defaultOffset = 5;
 		const triggerMetricsCollection = this._trigger.getClientRects();
 		const triggerMetrics = (triggerMetricsCollection.length > 0) ? triggerMetricsCollection[0] : {top: 0, left: 0};
 		const offset = {
 			top: triggerMetrics.top + distanceScrolled,
-			left: triggerMetrics.left
+			left: triggerMetrics.left,
+			right: defaultOffset
 		};
+		
 		const originalOffsetLeft = offset.left;
-
 		const arrow = this._content.querySelector(`.${this._arrowClass}`);
 		const windowWidth = document.documentElement.clientWidth;
 		const arrowHeight = 12;
 		const arrowWidth = 20;
-		const defaultOffsetLeft = 5;
-		const pagePadding = 32;
+		const pagePadding =  (this._options.PAGE_PADDING) ? this._options.PAGE_PADDING : 32;
 		const mdBreakPoint = 768;
 		const lgBreakPoint = 1024;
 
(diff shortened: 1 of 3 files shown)


tulipdexter:
Should there be additional tests to cover these changes?

toolkits/global/packages/global-popup/js/popup.js
```diff
@@ -95,22 +96,48 @@ const Popup = class {
 
 	_calcPositioning() {
 		const distanceScrolled = document.documentElement.scrollTop;
-		const triggerMetrics = this._trigger.getBoundingClientRect();
+		const triggerMetricsCollection = this._trigger.getClientRects();
```
tulipdexter, line 99: This is a jQuery method isn't it?

nickcall:
> Should there be additional tests to cover these changes?

Think your right was struggling with what they should be as a lot of the work is just altering what we had already. Will go away and come up with something

toolkits/global/packages/global-popup/js/popup.js
nickcall: https://developer.mozilla.org/en-US/docs/Web/API/Element/getClientRects

toolkits/global/packages/global-popup/js/popup.js
```diff
@@ -41,7 +41,8 @@ const Popup = class {
 		}
 		const pos = this._calcPositioning();
 		this._content.style.top = this._px(pos.top);
-		this._content.style.transform = `translateX(${pos.left}px)`;
+		this._content.style.left = this._px(pos.left);
+		this._content.style.right = this._px(5);
```
tulipdexter, line 45: Is 5 a magic number? Is it always the case that this is 5?

toolkits/global/packages/global-popup/js/popup.js
```diff
@@ -95,22 +96,48 @@ const Popup = class {
...
 		};
+		const originalOffsetLeft = offset.left;
+
 		const arrow = this._content.querySelector(`.${this._arrowClass}`);
 		const windowWidth = document.documentElement.clientWidth;
 		const arrowHeight = 12;
 		const arrowWidth = 20;
+		const defaultOffsetLeft = 5;
+		const pagePadding = 32;
+		const mdBreakPoint = 768;
```
tulipdexter, line 113: Should this be an option that is configurable when instantiating the class?

toolkits/global/packages/global-popup/js/popup.js
```diff
@@ -95,22 +96,48 @@ const Popup = class {
...
+		const lgBreakPoint = 1024;
```
tulipdexter, line 114: See above

nickcall: Yeah I'm not really sure they should be, breakpoints are breakpoints if they ever change they would have to be changed in the file passing them in or here and the breakpoints are the same across all imprints now I believe.

toolkits/global/packages/global-popup/js/popup.js
nickcall: 5 is indeed the magic number, De La Soul were wrong all along.

I'm going to make 5 the initial value and use this as a variable that can be passed into both the right and left by default. If it needs to change the code will do that. It's basically passed a left/right value to stretch the popup.

toolkits/global/packages/global-popup/js/popup.js
nickcall: I think your right as padding could be different going to make it configurable with a default value of 32 which is what it would be across our article pages.

tulipdexter (approved):

PR #475, opened 2021-02-25
Removes presentational styles from container utility

sky-jack:
This will be a breaking change for users of Springer Nature brand context that rely on the container utility for card/container styles.


context/brand-context/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@springernature/brand-context",
-  "version": "10.1.1",
+  "version": "10.1.2",
```
tulipdexter, line 3: Looks like you've bumped a patch but meant a major

tulipdexter (changes requested):

context/brand-context/package.json
sky-jack: Sorry. I should have removed that text from the PR. I discussed this with Alex and he suggested it should be a patch.

tulipdexter: Ah I see. In the case the `History` entry is incorrect 😄

context/brand-context/HISTORY.md
```diff
@@ -1,7 +1,10 @@
 # History
 
+## 11.0.0 (2021-02-25)
```
tulipdexter, line 3:
```suggestion
## 10.1.2 (2021-02-25)
```

tulipdexter (changes requested):

tulipdexter (approved):
Nice one @sky-jack!

alexkilgour:
This change was already approved but not merge. Due to another incoming change, this meant that the version was out of date. I have bumped the version and merged this PR to avoid this happening again.

PR #476, opened 2021-02-26
Remove archived components

alexkilgour:
Removes the following components:
- `global-article`, `nature-article`, `springer-article` which now live in Oscar
- `springer-lists` which has been moved to the global toolkit as `global-lists`
- `springer-pagination` which has been moved to the global toolkit as `global-pagination`

All the components are still available on NPM (no applications using these components will break) but have been marked as deprecated, linking to the details in the `DEPRECATED.md` file in this repository


tulipdexter (approved):

tulipdexter:
cashback

PR #477, opened 2021-03-01
Default font-sizing bug

alexkilgour:
Adds:
- `$context--font-size-sm`
- `$context--font-size-md`
- `$context--font-size-lg`

These variables exist in all context packages except `default` and are referenced by some global components. This meant that we get SASS compilation errors when trying to use the default brand.


saramorueta (approved):
LGTM 🙆‍♀️

PR #479, opened 2021-03-01
Capitalize first letter instead all words in links

ricky-lau:
Also update readme to reflect what is on live and remove two dropdown examples didnt show anything different.


saramorueta (approved):
LGTM

PR #480, opened 2021-03-01
Dont focus input

jon-stevens:
This PR removes a line which focuses the autocomplete input when you click outside of it. 
This bit of code is called when you click a suggestion (updates input value and removes suggestions), or when you click outside of the component (removes suggestions) - it probably doesn't need to focus on the input.

### DIFF
--- a/toolkits/global/packages/global-autocomplete/HISTORY.md
+++ b/toolkits/global/packages/global-autocomplete/HISTORY.md
@@ -1,6 +1,6 @@
 # History
 
-## 5.0.3 (2020-03-01)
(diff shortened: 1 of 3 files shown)


toolkits/global/packages/global-autocomplete/HISTORY.md
```diff
@@ -1,5 +1,8 @@
 # History
 
+## 5.0.3 (2020-03-01)
```
ricky-lau, line 3: year needs updating 🙈

ricky-lau (approved):

PR #483, opened 2021-03-03
Add an open method for popups to allow them to be lazily created when…

moddular:
… the trigger is first clicked

Creating the markup for author popups is a relatively expensive operation. We would like to be able to defer this until an author link is actually clicked instead of building them all up front. However if you simply create the popup object in the click event it won't open because it'll be too late for the popup itself to respond to the event. 

This change adds an open method to popup objects so we can programatically open the popup after it has been created. 

e.g.

```javascript
trigger.addEventListener('click', function() {
    const popup = new Popup(trigger, 'popupcontent');
    popup.open();
}, {capture: false, once: true});
```

### DIFF
--- a/toolkits/global/packages/global-popup/HISTORY.md
+++ b/toolkits/global/packages/global-popup/HISTORY.md
@@ -1,5 +1,8 @@
 # History
 
+## 4.3.0 (2021-03-03)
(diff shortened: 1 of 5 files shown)


benjclark (approved):
Nice one 👍

PR #486, opened 2021-03-11
Remove save data variable and update save data fonts

nickcall:


benjclark (approved):
Yeah I think it's good argument to make re breaking change. Given that it's brand context, people might be using that save data variable in their code

PR #491, opened 2021-03-17
Small refactor to generate a smaller compiled bundle size

moddular:
This change seems to allow us to remove:

```
import 'core-js/features/dom-collections';
import 'core-js/features/symbol';
import 'regenerator-runtime/runtime';
import 'core-js/stable/object/assign';
```

from header-150.js and would take the resulting bundle size from around 32kb to 7kb.

### DIFF
--- a/toolkits/nature/packages/nature-header/HISTORY.md
+++ b/toolkits/nature/packages/nature-header/HISTORY.md
@@ -1,5 +1,8 @@
 # History
 
+## 5.1.3 (2021-03-17)
+    * Refactor to generate smaller bundle size
+
 ## 5.1.2 (2021-03-01)
     * Capitalize first letter instead all words in link
     * Update readme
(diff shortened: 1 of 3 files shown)


tulipdexter:
Haha that is insane!

toolkits/nature/packages/nature-header/HISTORY.md
```diff
@@ -1,5 +1,8 @@
 # History
 
+## 5.2.0 (2021-03-17)
```
tulipdexter, line 3: Not just a patch?

ricky-lau (approved):
this is great 🚀

toolkits/nature/packages/nature-header/HISTORY.md
moddular: Good point - probably should be - I'll update it

tulipdexter (approved):

PR #494, opened 2021-03-29
conditional focussing

jon-stevens:
I recently removed some code which returned focus to the input after selecting/cancelling a suggestion.
This PR reintroduces the focussing but only in the right conditions.
If you click/select a suggestion, then return focus to the input, but if you click outside of the input/suggestions then don't.
Also, if you press Escape the list of suggestions will dissapear and focus is returned to the input.

I will clean up the test in a further PR (remove duplication etc).

https://user-images.githubusercontent.com/11961647/112858754-140a6e00-90aa-11eb-99e7-19179eb780dc.mov


moddular (approved):
:+1:

alexkilgour (approved):
👌

PR #503, opened 2021-04-19
Fluid typography on titles of nature-hero and global-card

ricky-lau:
updates titles on nature-hero and global-card components to use fluid typography set from https://github.com/springernature/frontend-toolkits/pull/497


benjclark (approved):

tulipdexter:
Just curious, if it's using fluid typography, why use a breakpoint?

ricky-lau:
that was for the nature-hero when the container changes from full width to the text and image beside each other at `md` . The h1 was too tight in the space and was tweaked to be smaller.

PR #505, opened 2021-04-19
Attempt a fix for hasAttribute error we're seeing in sentry

moddular:


tulipdexter (approved):
Curious to know what browsers are getting through CTM to get the JS but then throwing up issues about `hasAttribute` 🤔  ✨ 

Didn't know about the `hash` property. Everyday's a school day!

moddular:
It's odd, it happens in seemingly modern browsers, but I've never been able to reproduce it. It's like sometimes the node list contains something that isn't a node.

PR #525, opened 2021-06-10
Add border to global-pill to increase contrast

ricky-lau:
* Add default settings and to nature settings
* Add overrides for margin values to nature settings
* Add border to increase contrast from background
* Update branding info and example in readme

### DIFF
--- a/toolkits/global/packages/global-pill/scss/10-settings/nature.scss
+++ b/toolkits/global/packages/global-pill/scss/10-settings/nature.scss
@@ -1,2 +1,12 @@
+/**
+ * @springernature/global-pill
+ * Nature skin settings
+ */
+
+// Default setting
+@import 'default';
+
 $pill--background-color: color('light-blue');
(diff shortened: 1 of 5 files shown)


toolkits/global/packages/global-pill/scss/10-settings/nature.scss
```diff
@@ -1,2 +1,12 @@
...
+ * @springernature/global-pill
+ * Nature skin settings
+ */
+
+// Default setting
+@import 'default';
+
 $pill--background-color: color('light-blue');
 $pill--color: color('action-base');
+$pill--margin-right: spacing(4);
```
benjclark, line 11: should these variable be called within the component scss?

benjclark: oh they are - it's just that the diff stopped at line 7, silly me

benjclark (approved):
Nice

PR #526, opened 2021-06-18
Fixes for explorer!

jon-stevens:
Explorer fixes:

Use 'Esc' as well as 'Escape' keycode.
Use .removeChild() rather than .remove().


benjclark (approved):
👍

tulipdexter (approved):

PR #532, opened 2021-07-13
[NM] SN new skipLink background color

nasiamakryg:
Change sn background color for skip link to follow new palette because of new brand context


alexkilgour (approved):
👌

PR #534, opened 2021-07-14
[NM] fix link colors

nasiamakryg:


context/brand-context/springernature/scss/40-base/basic.scss
```diff
@@ -11,7 +11,7 @@ body {
 
 a {
 	text-decoration: none;
-	color: color('blue');
+	color: color('medium-blue');
```
alexkilgour, line 14: This can be `$context--link-color` I would think. It's available to this file

alexkilgour (approved):
🌈

PR #538, opened 2021-07-28
Removes u-separator utility styles 

benjclark:
I have performed a search across all SN repos and this Springer utility classname is not being used anywhere.

I also intend to replace this with a shared mixin/utility classname for separator styling


saramorueta (approved):
LGTM

PR #554, opened 2021-08-12
Add focus style for tabindex and contenteditable attributes in nature

saramorueta:
Screenshot 2021-08-12 at 11 19 24

### DIFF
--- a/context/brand-context/nature/scss/40-base/layout.scss
+++ b/context/brand-context/nature/scss/40-base/layout.scss
@@ -58,7 +58,5 @@ abbr[title] {
 
 [contenteditable],
 [tabindex="0"] {
-	&:focus {
-		outline: 4px solid color('action-focus');
-	}
+	@include u-focus-outline();
 }
--- a/context/brand-context/nature/scss/40-base/layout.scss
+++ b/context/brand-context/nature/scss/40-base/layout.scss
@@ -55,3 +55,10 @@ small {
 abbr[title] {
 	text-decoration: none;
 }
+
+[contenteditable],
+[tabindex="0"] {
+	&:focus {
+		outline: 4px solid color('action-focus');
+	}
+}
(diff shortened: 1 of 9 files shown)


context/brand-context/nature/scss/40-base/layout.scss
```diff
@@ -55,3 +55,10 @@ small {
 abbr[title] {
 	text-decoration: none;
 }
+
```
ricky-lau, line 58: Noticed that this is slightly different from the one we have in `40-base/links.scss`. color Do they need to be different?

saramorueta: I know what you mean, but the value for both is the same than we use for the links 😆 https://github.com/springernature/frontend-toolkits/blob/6b150f6101095e687c632371afd8dc0df5b338ab/context/brand-context/nature/scss/10-settings/colors/default.scss#L27 

do you think is better to use the`color('lemon')` instead of adding a new `action-focus` to the map?

ricky-lau: You have for outline is 4px instead of 3px as well.. why not have the extra selectors added to the existing focus style in `links.scss`? I suppose what you are adding don't always belong to links 🤔 maybe this could be a mixin that gets included so focus style stays consistent in one place.

Regarding adding the action-focus to map. I don't have a strong opinion about it. We rarely use the lemon color for anything other than focus color so maybe it'll add context to that color to not use for anything else 😁

saramorueta: I added a couple of changes following your great suggestions! 

the variable called `$context--focus-color` existed already for the springernature brand-context, so I guess it makes sense to override it by each brand after all?

please let me know your thoughts!

ricky-lau (approved):
🥳

ricky-lau (approved):

saramorueta:
thanks for reviewing @ricky-lau great suggestions!! you rule!

PR #556, opened 2021-08-13
Add parameter for reflow on focus mixin 

ricky-lau:
* Add parameter for reflow on focus mixin
        - applied on links as this seems to be where outline appears incorrectly
        - related to 13.1.2
        
* Fix `:focus` being applied twice on `a`

### DIFF
--- a/context/brand-context/default/scss/30-mixins/focus.scss
+++ b/context/brand-context/default/scss/30-mixins/focus.scss
@@ -1,6 +1,8 @@
-@mixin u-focus-outline {
+@mixin u-focus-outline($reflow: false) {
 	&:focus {
 		outline: 3px solid $context--focus-color;
-		-webkit-font-smoothing: antialiased; // reflow needed to ensure outline styles visible in Safari v13+
+		@if $reflow == true {
+			-webkit-transform: translateZ(0); // reflow needed to ensure outline styles visible in Safari v13+
+		}
 	}
 }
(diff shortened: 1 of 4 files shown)


context/brand-context/default/scss/30-mixins/focus.scss
```diff
@@ -1,6 +1,8 @@
-@mixin u-focus-outline {
+@mixin u-focus-outline($reflow: false) {
 	&:focus {
 		outline: 3px solid $context--focus-color;
-		-webkit-transform: translateZ(0); // reflow needed to ensure outline styles visible in Safari v13+
+		@if $reflow == true {
+			-webkit-transform: translateZ(0); // reflow needed to ensure outline styles visible in Safari v13+
```
alexkilgour, line 5: Is there a negative effect of applying this all browsers? i.e. just using `transform: translateZ(0)`

ricky-lau: Replaced the `transform: translateZ(0)` hack with `will-change: transform` to create new layer. 

Removed the parameter and applied it to all browsers as I couldn't see significant effects on the page when tabbing through the page; the tabbing isn't a smooth transition that can be noticed.

context/brand-context/default/scss/30-mixins/focus.scss
```diff
@@ -1,6 +1,6 @@
-@mixin u-focus-outline {
+@mixin u-focus-outline($reflow: false) {
```
alexkilgour, line 1: can remove `$reflow: false` from here?

context/brand-context/nature/scss/40-base/links.scss
```diff
@@ -4,6 +4,7 @@
  */
 
 a {
+	@include u-focus-outline($reflow:true);
```
alexkilgour, line 7: can remove `$reflow: true` from here?

alexkilgour (approved):

PR #557, opened 2021-08-18
Remove flex from sm breakpoint

jon-stevens:
Remove flex from authors list at small breakpoint.
Remove it entirely from metadata row at top as they're all inline anyway.

Screenshot 2021-08-18 at 13 26 43


alexkilgour (approved):
👌

PR #559, opened 2021-08-24
add springer brand to global-skip-link

ricky-lau:
- add springer brand in settings to use for springer and bmc/so
- bump brand-context to the latest


benjclark (approved):
nice one

PR #561, opened 2021-09-01
add global-facet package

ricky-lau:
moves over what was used on nature search pages in oscar


toolkits/nature/packages/nature-facet/scss/50-components/facet.scss
```diff
@@ -0,0 +1,171 @@
...
+	font-weight: $context--font-weight-normal;
+	margin-bottom: 0;
+	padding: spacing(8);
+	white-space: normal;
+	width: 100%;
+
+	@include media-query('sm') {
+		margin-bottom: 0;
+		white-space: nowrap;
+		width: 111px; // hardcoded to fit 4 items in a row
```
benjclark, line 78: Do these hardcoded width values assume the consumer of this component has implemented anything in particular? e.g. u-container must be used on pages that have this component?

toolkits/nature/packages/nature-facet/js/enhanced-facet.js
```diff
@@ -0,0 +1,114 @@
...
+
+const classNames = {
+	SELECTED: 'c-facet__selected',
+	ELLIPSIS: 'c-facet__ellipsis',
+	CLEAR: 'c-facet__clear-selection',
+	BUTTON_CONTAINER: 'c-facet__button-container',
+	SUBMIT: 'c-facet__submit'
+};
+
+const buttonName = {
```
matteo-gildone, line 23: Maybe buttonCopy?

toolkits/nature/packages/nature-facet/scss/50-components/facet.scss
```diff
@@ -0,0 +1,171 @@
...
+		text-decoration: none;
+
+		svg {
+			transform: rotate(180deg);
+		}
+	}
+}
+
+.c-facet__button--disabled {
+	border-color: #c7c4bf;
```
benjclark, line 104: Should this colour be variablised because it's possible that other brands will want to overwrite the disabled colour?

toolkits/nature/packages/nature-facet/js/enhanced-facet.js
```diff
@@ -0,0 +1,114 @@
...
+
+	button.addEventListener('click', event => {
+		event.preventDefault();
+		const checkedInputs = element.querySelectorAll('input[type=checkbox]:checked, input[type=radio]:checked');
+
+		checkedInputs.forEach(function (checkedInput) {
+			checkedInput.checked = false;
+		});
+
+		button.closest('form').submit();
```
matteo-gildone, line 49: `closest` polyfill?

toolkits/nature/packages/nature-facet/scss/50-components/facet.scss
```diff
@@ -0,0 +1,171 @@
+.c-facet {
+	background-color: #f3f3f3;
```
benjclark, line 2: Should we create a variable to allow consumers to change background colour relevant to their brand? If this is a grey, should we use the greyscale function in toolkits?

toolkits/nature/packages/nature-facet/scss/50-components/facet.scss
ricky-lau: I'm thinking to refactor these buttons to take up the full width of a container to replace the hardcoded width value. Also within a flexbox container as the intention was to have the buttons on one row unless narrow view where it is stacked. Then have the container widths managed outside of this package as the number of buttons could vary depending on the page if we don't want the buttons to be too wide.

toolkits/nature/packages/nature-facet/scss/50-components/facet.scss
ricky-lau: I'll see if there is an existing colour for this and the comment below in the toolkit otherwise I'll put it a variable

toolkits/nature/packages/nature-facet/scss/50-components/facet.scss
benjclark: Sounds good.

So sorry, I only realised after leaving this comment that this is a Nature scoped package and not Global. But if you do remove the hardcoded values and let those be managed outside of the package, as you suggest, then could this become a Global component instead?

toolkits/nature/packages/nature-facet/js/enhanced-facet.js
ricky-lau: would you be persuaded with `buttonText`? `buttonCopy` sounds a bit like the button is being or has been cloned

toolkits/nature/packages/nature-facet/js/enhanced-facet.js
ricky-lau: Was trying to find out if the IE11 polyfill bundle contains `closest` but found we might have an issue with the recent polyfill changes. I've reached out to Andy to see if he can help.

toolkits/nature/packages/nature-facet/js/enhanced-facet.js
matteo-gildone: I think `buttonText` is amazing!

toolkits/nature/packages/nature-facet/js/enhanced-facet.js
ricky-lau: ive added a note in the readme that this requires polyfill

toolkits/nature/packages/nature-facet/scss/50-components/facet.scss
ricky-lau: This is living as variables in settings. Did see that we have disabled button theme but it is not suitable for the nature brand since it is not on a white background.

toolkits/global/packages/global-facet/scss/10-settings/default.scss
```diff
@@ -0,0 +1,33 @@
...
+ * Facet
+ * Default branding for component
+ */
+
+$facet--background-color: lightgrey;
+$facet--color: black;
+$facet--font-bold: 700;
+$facet--font-size-small: 1.4rem;
+$facet--font-size: 1.6rem;
+$facet--spacing-x-small: 4px;
```
benjclark, line 11: We have a default spacing map in brand context so it's probably worth defining these using the spacing function .e.g. `spacing(4)`

```$context--spacing: (0, 2, 4, 8, 16, 24, 32, 48, 64);```

File: toolkits/global/packages/global-facet/README.md [hunk 7]
```diff
@@ -0,0 +1,87 @@
...
+```html
+<div class="c-facet" data-facet>
+    <!-- example of checkbox input -->
+    <div class="c-facet__container">
+        <fieldset class="c-facet__item">
+            <legend>
+                <span class="c-facet__label">Subject</span>
+                <span class="u-hide u-js-show">
+                    <button type="button" class="c-facet__button c-facet__button--border" data-facet-expander data-facet-target="#subject-target">
+                        <svg class="c-facet__icon" role="img" aria-hidden="true" focusable="false" height="16"
```
benjclark, line 30: I think we can remove the width and height attributes from the example <svg> here, as per the outcomes of the discussions Heydon Pickering started the other day. https://springernature.slack.com/archives/C0GDUEGMS/p1631007607070800

benjclark: Oh scratch that - I've just seen that you are including the `u-icon` mixin within the facet scss 👍

benjclark (approved):
Thanks for turning this in to a Global Component ❤️

benjclark (approved):

PR #562, opened 2021-09-07
Fit icons to surrounding text as a default

Heydon:
## 1. Removal of `height` and `width` attributes on source icons

Source SVG files should only define the `viewBox` coordinate system. 

Where consumers size an `<svg>` containing a `<use>` element for the icon source, the two sets of `width` and `height` attributes (or source `width` and `height` and CSS) work against each other and cropping occurs. Removing `width` and `height` from the source files prevents this and allows adjustment of sizing without cropping.

## 2. Addition of `em`-based sizing

The `.u-icon` utility now includes a `width` and `height` of `1em` to size the icon to fit the surrounding text size by default. Font metrics can mean `1em` isn't necessarily an optimal value but this has been tested against Merriweather and looks decent.

Consumers/brands appear to currently be using the provided `width` and `height` values (which differ between icons) on the template `<svg>`. In one case, the toolkit SVG (search) has been redrawn as a new icon using a 16X16 coordinate coordinate system (instead of 22X22), apparently just to emulate `16px` (`1em` for body text). Such things will no longer be necessary.

## 3. Breaking change so major version bump

`14.0.0` -> `15.0.0`


alexkilgour (approved):
Perfect 👌

alexkilgour (approved):
👍

PR #564, opened 2021-09-09
add nature headings component

matteo-gildone:
## Nature Headings component 
Screenshot 2021-09-09 at 11 30 56

A couple of questions:
Should it be a utility class inside Nature context instead of a component?
If stays as a component should I keep the mixin?


toolkits/nature/packages/nature-headings/scss/10-settings/headings.scss
```diff
@@ -0,0 +1,5 @@
+$headings--border-top: 5px solid greyscale();
```
ricky-lau, line 1: Change `greyscale()` to `black` to be less ambiguous?

ricky-lau:
I think keep it as a component as its doing a few things to be a utility 🤔 
The mixin then isn't worthwhile having as it's only being used once and its an extra file you have to add to the entry points

ricky-lau (approved):

PR #565, opened 2021-09-09
Use commas in template rather than scss

jon-stevens:
Remove the slightly complicated addition of comma separators in the SCSS. Add them in the template instead.
And move the author background to a variable.

Screenshot 2021-09-09 at 11 32 15


benjclark (approved):
giphy

PR #567, opened 2021-09-14
chore: bring linting up to date

Heydon:
* Updated dependencies/versions
* Gone through and fixed all the errors

There are a number of errors in tests but I think they were already present since they don't seem to be related to the linting fixes. Would be great to get a second opinion.


tulipdexter:
As a discipline we still have to support IE (technically 11 but our cutting the mustard technique lets through 10 as well so 🙈 ) and so a lot of the things linting throws up we have to end up adding an override in the springernature eslint config. E.g. `Element.remove()` and `Element.append()`.

Heydon:
@tulipdexter 👋 Makes sense, but I'm using SN's rules? Shouldn't the issues with remove() and append() be taken care of my the config itself?

tulipdexter:
@Heydon yeah you're absolutely right. I'm not sure why they are not set in the config 🤔

tulipdexter:
https://github.com/springernature/eslint-config-springernature/blob/master/configurations/core.js#L279 🤔 🤔

Heydon:
@tulipdexter Huh! 

Do my local deps look right? I ran `npm ci` to get the new config pulled in locally.

```
    "@babel/plugin-transform-runtime": "^7.5.0",
    "@babel/preset-env": "^7.6.3",
    "@babel/runtime": "^7.6.3",
    "@springernature/eslint-config": "^5.0.0",
    "@springernature/frontend-package-manager": "^7.1.0",
    "@springernature/sasslint-config": "^1.2.1",
    "@springernature/util-eslint-runner": "^0.0.3",
    "@springernature/util-sass-compiler": "^0.1.0",
    "acorn": "^7.1.0",
    "eslint": "^7.17.0",
    "eslint-plugin-import": "^2.22.1",
    "eslint-plugin-jest": "^24.1.3",
    "eslint-plugin-no-use-extend-native": "^0.5.0",
    "eslint-plugin-node": "^11.1.0",
    "eslint-plugin-promise": "^4.3.1",
    "eslint-plugin-unicorn": "^28.0.1",
    "jest": "^26.4.2",
    "lerna": "^3.18.3",
    "sass-lint": "^1.13.1"
```

Heydon:
Or maybe there's something up with my `.eslintrc`? I note the repo' is eslint-config-springernature, not eslint-config, but the package I pull in is eslint-config, with the 5.0.0 version number 🤔

```
{
	"extends": [
		"@springernature/eslint-config",
		"@springernature/eslint-config/node",
		"@springernature/eslint-config/jest"
	],
	"parserOptions": {
		"ecmaVersion": 9
	},
	"rules": {
		"import/no-unresolved": "off"
	}
}
```

Heydon:
Okay, my local core file has that line too 🤔🤔🤔

tulipdexter:
The npm package is `@springernature/eslint-config` so that sounds right. I'm not sure about the repo naming, I suspect it existed before the `@` stuff but @alexkilgour will know more about that.

tulipdexter:
It does have a fair amount of peer dependencies, did you use `npx`?

tulipdexter:
I don't know if this helps you at all?
https://github.com/springernature/oscar/blob/master/lib/frontend/.eslintrc

Heydon:
@tulipdexter Thanks for checking back on this! I ran `npx` for peer deps then tried again, changing a `remove` to `parentNode.removeChild` and... it still doesn't like it 🤔

tulipdexter:
@Heydon and I took this conversation further in Slack, but for documentation purposes:

We believe the problem is that the bumping of the `@springernature/eslint-config` version meant that peer dependencies were bumped, and as a result of a `major` release to unicorn plugins, some [rule names had been changed](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/deprecated-rules.md). This meant that the use of `appendChild()` was being flagged by the linter because the rule we had set to switch it off in our config was using an old name so had no effect.

PR #572, opened 2021-09-20
add flexbox feature query on nature-hero 

ricky-lau:
iOS 8 (and other older browsers requiring `display: -webkit-flex`, `-webkit-flex-wrap` not added by oscar's [browserslist](https://github.com/springernature/oscar/blob/master/lib/frontend/.browserslistrc) ) was not happy with nature-hero using flexbox:

image

Design of hero reverts to the 'mobile view' without flexbox:

image


toolkits/nature/packages/nature-hero/HISTORY.md
```diff
@@ -1,5 +1,8 @@
 # History
 
+## 1.1.1 (2021-09-29)
```
jon-stevens, line 3:
```suggestion
## 1.1.1 (2021-09-21)
```

PR #573, opened 2021-09-21
Accommodate diff sized logos at breakpoints

jon-stevens:
We try to accommodate different sized logos at different breakpoints. One logo in particular with a lot of whitespace at the end has illustrated a bit of a problem whereby it gets shrunk quite small at smaller breakpoints. 
Below you can see the green-bordered SVG which seems to have been set to have ~40% whitespace?

Screenshot 2021-09-21 at 11 57 51

--->

Screenshot 2021-09-21 at 11 44 32

Screenshot 2021-09-21 at 11 44 19

### DIFF
--- a/toolkits/springernature/packages/springernature-header/HISTORY.md
+++ b/toolkits/springernature/packages/springernature-header/HISTORY.md
@@ -1,5 +1,8 @@
 # History
 
+## 2.2.0 (2021-09-21)
+    * Try to accommodate for different sized journal logos at different breakpoints
+
 ## 2.1.0 (2021-09-15)
     * Allow journal header to take a little more room
 
--- a/toolkits/springernature/packages/springernature-header/package.json
+++ b/toolkits/springernature/packages/springernature-header/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@springernature/springernature-header",
-  "version": "2.1.0",
+  "version": "2.2.0",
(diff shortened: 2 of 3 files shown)


sturobson:
Is it possible to improve the source file and remove the whitespace from the SVG?

jon-stevens:
> Is it possible to improve the source file and remove the whitespace from the SVG?

Hey. I would like to! The journal logos come from a branding service. I am not sure where that is actually actually located.
Given the number of journals it is always a possibility that there will be images like this so it's a good idea to try and cater for that. I am going to try and find if I can fix the image file though.

ricky-lau:
I wonder if the logo you have is shared with the one on the journal sites? 
https://www.nature.com/npjmentalhealth/ - this also seems to have the same issue with the extra space.

They get added on pubserv:
https://product-admin.nature.com/admin/products/npjmentalhealth

jon-stevens:
> They get added on pubserv:
> https://product-admin.nature.com/admin/products/npjmentalhealth

Hmm do you know where I could request access?

ricky-lau:
I shall find out! 🔍

ricky-lau:
Turns out I had the power to give access and we looked at editing the logo to remove extra whitespace and made sure it looked ok for the journal site and editorial system.

ricky-lau (approved):

PR #576, opened 2021-09-23
Improve logo sizing for non-js view

jon-stevens:
After fixing the logo sizing in a previous commit, I noticed it was still a bit squiffy when js disabled.
This fix basically adjust the widths/floating for non-js.

### DIFF
--- a/toolkits/springernature/packages/springernature-header/scss/50-components/enhanced.scss
+++ b/toolkits/springernature/packages/springernature-header/scss/50-components/enhanced.scss
@@ -1,58 +1,64 @@
-.js {
-	.c-header__content {
-		position: relative;
+.c-header__content {
+	position: relative;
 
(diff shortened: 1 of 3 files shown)


toolkits/springernature/packages/springernature-header/scss/50-components/enhanced.scss
```diff
@@ -1,58 +1,64 @@
...
-		flex-wrap: wrap;
-		align-items: center;
+.c-header__content-block {
+	display: flex;
+	flex-wrap: wrap;
+}
 
+.c-header__logo-link {
+	display: flex;
+	flex-basis: 100%;
```
ricky-lau, line 49: flex-basis is here twice

ricky-lau (approved):

PR #577, opened 2021-09-27
Layout with sidebar

Heydon:
The first of a number of possible global layout primitives: generic layout components to help with composition. 

Created to address the need for a 2-column / sidebar layout for springer.com, but potentially useful in numerous scenarios. 

Works without media queries, making it context independent.

sidebar


tulipdexter:
Just wondering why you chose `c-with-sidebar` rather than, say, `c-layout-with-sidebar`?

tulipdexter:
I didn't realise `var()` could take a fallback, that's neat.

Heydon:
> Just wondering why you chose `c-with-sidebar` rather than, say, `c-layout-with-sidebar`?

@tulipdexter Oh, good catch. I've been inconsistent there! Will amend.

sturobson:
> Just wondering why you chose c-with-sidebar rather than, say, c-layout-with-sidebar?

How about `l-with-sidebar` which would set a new precedent for more primitive layout components that could be additive to `c-components` 

```
<article class="l-with sidebar | c-article-teaser">
…
</article>
```

Heydon:
@sturobson I'm on board with that. I think `c-` denotes a lexical component, as in something with actual functionality, while `l-` relates to lower level helpers?

tulipdexter:
Yeah I'm also on board with that, but I don't think `l` would be lower level helpers. I think it's explicitly layout.

tulipdexter:
@alexkilgour do you have a view on this?

Heydon:
@tulipdexter Sorry, yeah, I meant helper with a little 'h', just trying to describe the purpose of it. Cool, I'll change that.

Heydon:
Hmm. Do I change the package name to just `global-with-sidebar` now though?

Heydon:
@tulipdexter Am I right in saying the classnames absolutely have to match the package name (minus the `c-` or `l-` prefix)? Because I think having `layout` in the package name would be good, for each layout (so still `global-layout-with-sidebar`) but it would be redundant to repeat it in the classname like `l-layout-with-sidebar`... Or maybe there's a better way?

tulipdexter:
@Heydon I think I'd be incline to have it as a grouping of layouts. My assumption is that there could be more layouts than just this one? So something like `global-layout`. A bit like this https://github.com/springernature/frontend-toolkits/tree/master/toolkits/global/packages/global-lists.

Then again, I'm not sure how it works with the CSS folder structure, i.e. `50-components`?

Heydon:
@tulipdexter I like that idea, except we're also trying to make packages consumable (hence all the `.hbs` files I've merged in recently). There's no way to currently expose multiple templates/models and the demo script still work (I don't think?).

tulipdexter:
Hmmmm. Do we have any rough idea how much work it is to make that work? I feel like we will inevitably need this at some point, unless we want to split things like `lists` into singular packages.

Heydon:
> Hmmmm. Do we have any rough idea how much work it is to make that work? I feel like we will inevitably need this at some point, unless we want to split things like `lists` into singular packages.

True. I think the layouts being in separate packages would make sense otherwise people will end up importing redundant CSS etc. But you're right, it would be useful to have multiple `.hbs` files per package (for the lists, for example). @alexkilgour might be able to help.

benjclark:
So what's your stance on IE11 support? As I guess `gap` and `var()` are unsupported.

Hopefully the current discussions about dropping support for IE11 conclude soon in favour of dropping it? 🤞 😄

Heydon:
@benjclark Yes, gap won't work which will mean layouts are "flush" 😬 but content will still be available to the core browsers. With the cutting the mustard technique, I could write in margins for IE10/11, though? And enhance to gap for other browsers?

benjclark:
Perhaps this is a case for `@supports` ? with a margin-based fallback inside a `@supports not`?

It's annoying. But until the IE11 re-classification discussion is resolved it's probably the correct thing to do? 🤷‍♂️ 🤷 

Also, if you and Stu and have any thoughts for or against or rational on the IE11 re-classification then feel free to share in the ie11-core-browser-analysis channel

Heydon:
> Perhaps this is a case for @supports ? with a margin-based fallback inside a @supports not?

@benjclark I'm afraid not, since there is no way to differentiate between `gap` for Grid and gap for Flexbox 😩. Bit of an unfortunate one, standards wise.

benjclark:
@Heydon Maybe it doesn't matter because both are unsupported by IE11 and lets face it with Microsoft announcing it's ending support for IE11 on Win 10 next year I think there's little to no chance they are going to implement support for either of those at this point? 

What do you think? or am I being too reckless?

sturobson:
Hmm, I think we _might_ be able to target IE11, but this would effect some very old versions of evergreen browsers (Edge, Chrome, Firefox, Safari).

Looking at the caniuse IE11 does not understand `grid-gap` and therefor `gap` … autoprefixer fixes this by adding extra columns and rows (yeeesh!)

So to create something that IE11 would understand but have gap for the evergreen browsers we could be able to do.

```css
.thing {
  margin-right: 24px;

  @supports (gap: 24px;) {
    margin-right: 0; 
    gap: 24px;
  }
}
```

toolkits/global/packages/global-layout-with-sidebar/README.md
```diff
@@ -0,0 +1,67 @@
...
+```
+
+## Usage
+
+### Default settings
+
+Using the default settings, the `c-with-sidebar` element will be `400px` in the 2-column configuration. Both columns grow to 100% width of their parent `c-with-sidebar` where they wrap into 1 column. By default, this happens when the non-sidebar (“Main content” below) becomes less than `50%` of the total width.
+
+```html
+<div class="c-with-sidebar">
```
benjclark, line 27: The classnames in this readme need updating to the new l- prefix

benjclark:
@Heydon 
Nice! I just tested in IE11 on Win 10 and IE11 on Win 8.1 using our springer.com journal pages as a test subject to see it working and also to see how the existing site behaves if we were to start using it for layout styling. Looks good, can't see any issues. 

Should hopefully be able to get you both access to our Browserstack account if it helps at all.

The springer.com journal pages currently use a 1024px breakpoint to control the sidebar collapse and the sidebar on that page is currently fixed at `454px`. I figured in order to achieve the same behaviour then I would simply configure `$with-sidebar--basis` to be `454px` and configure `$with-sidebar--min` to be `53%` ((breakpoint - sidebar width - gutters)/breakpoint * 100 which is (1024-454-32)/1024*100) which seems to work well.

output4

toolkits/global/packages/global-layout-with-sidebar/README.md
Heydon: @benjclark WHOOPS. Yep, very true. Will quickly do that now.

toolkits/global/packages/global-layout-with-sidebar/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "@springernature/global-layout-with-sidebar",
+  "version": "0.0.0",
```
benjclark, line 3: Think this needs to be 0.0.1 to match what you have written in history.md

sturobson:
❗️aside:

Just a thought -- is there a set naming convention for Sass variables? I cannot find it in the front-end playbook. 

I'd assume there's some general conventions that need to be adhered to so we don't get to possible duplication of Sass variables across project?

benjclark:
>Just a thought -- is there a set naming convention for Sass variables? I cannot find it in the front-end playbook.
I'd assume there's some general conventions that need to be adhered to so we don't get to possible duplication of Sass variables across project?

Yeah, good point. The Sass variable naming convention is something that evolved over time I think, perhaps on Slack chat only. I'll open a PR to add it to the how-we-write-css.md, maybe after the HTML class naming section?

Heydon:
@sturobson @benjclark I'm following the conventions found in the other toolkits/packages of prefixing with the component name and adding `--`. I'm not sure if there are hypothetical restrictions over what can be placed after `--`, though. I just chose what felt clear.

Heydon:
@benjclark I think I need to change the HISTORY again for publishing? So I did that too. Good to go?

benjclark (approved):
sadsadad

benjclark:
>So there could be:
u for utility things
c for components that do things
l for primitive layout

Off the back of this I guess we need to update our docs to define the new primitive layout prefix. I guess this is the place: https://github.com/springernature/frontend-playbook/blob/2673332cfef6e5852418dd8b95ad778c68f5b2f6/css/how-we-write-css.md#namespaces

PR #582, opened 2021-10-06
xsmall and tertiary button for SN use

nasiamakryg:
IN researcher portal we need a button which is not so loud and @foxintherain came up with this design which the recommended changes implement: 
Screen Shot 2021-10-06 at 15 16 29

### DIFF
--- a/context/brand-context/package.json
+++ b/context/brand-context/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@springernature/brand-context",
-  "version": "15.1.0",
+  "version": "16.0.0",
   "license": "MIT",
   "description": "Bootstrapping for all components and products",
   "keywords": [],


context/brand-context/HISTORY.md
```diff
@@ -1,4 +1,7 @@
 # History
+## 15.1.0 (2021-10-06)
+    * Add default setting, mixin and utility class for xsmall & tertiary button
```
alexkilgour, line 3: use the format `FEATURE: Add default setting...`

context/brand-context/HISTORY.md
```diff
@@ -1,4 +1,10 @@
 # History
+## 16.0.0 (2021-10-06)
+    * BREAKING:
+        * Update the plus icon
+    * Add a minus svg icon
```
alexkilgour, line 5: This is actually breaking as well because it is `remove icon-minus.svg, add minus.svg`

foxintherain:
NOICE ONE

PR #610, opened 2021-11-04
Adds springernature favicons

aliuk2012:
Historically these were provided to each team from the teams designer
and there was no central repository for these favicons. This would make
it extremely hard to find/replace all instances of these favicons across
the organisation.


alexkilgour:
You can release a new version of the context with these icons in the same PR. Just update the `HISTORY.md` file and the `package.json` with a new minor version (for new feature as per semver) and detail in the history file what the new feature is and for which brand. This will then auto-publish after a successful merge to the base branch

alexkilgour (approved):
perfect 👌

PR #611, opened 2021-11-05
Footer landmark

Heydon:
Removes `<footer>` from the `springer-publisher-footer` template to avoid nested footers when combined with `global-corporate-footer`.


tulipdexter:
Seems reasonable to me, but is it a breaking change?

alexkilgour (approved):
Make sense 👍

Heydon:
@tulipdexter I don't think so, because I don't think anyone but me is consuming the demo templates directly yet 😅

PR #620, opened 2021-11-12
Add optional width and height to set for u-icon mixin

ricky-lau:
Adding the option to change size as the default `1em` made some existing icons smaller than before.


saramorueta (approved):
fantastic! thank you!

PR #622, opened 2021-11-18
Allow default tab behaviour

jon-stevens:
This fix allows the default expected behaviour from the Tab keypress, i.e move focus to next focussable element, rather than scroll through the results list.


toolkits/global/packages/global-autocomplete/__tests__/index.js
```diff
@@ -1,219 +1,253 @@
 'use strict';
```
jon-stevens, line 1: The changes in the test here are largely just rejigging, removing duplication etc. I've added a new test block from line 193.

alexkilgour (approved):
Makes sense 👌

PR #623, opened 2021-11-22
Cluster layout

Heydon:
From the README:

_The Cluster component is a simple layout primitive for grouping (_clustering_) elements together. It improves upon basic `inline` and `inline-block` solutions by using Flexbox and the `gap` property to apportion space only _between_ the child elements._

Note that the `gap` is added progressively. Margin is used where it isn't supported.


benjclark (approved):

toolkits/global/packages/global-layout-cluster/scss/50-components/layout-cluster.scss
```diff
@@ -0,0 +1,36 @@
...
+.l-cluster > * {
+	/* ↓ Because margin doubles up */
+	margin: calc(#{$cluster--gap} / 2);
+	@supports (gap: 1em) {
+		margin: 0;
+		gap: var(--cluster--gap, #{$cluster--gap});
+	}
+}
+
+.l-cluster--right {
```
sturobson, line 17: I know `justify-content` doesn't make use of logical properties. But I wonder if moving toward that naming convention would be a good idea.

So, for `l-cluster--right` there would be `l-cluster--inline-end`.

toolkits/global/packages/global-layout-cluster/scss/50-components/layout-cluster.scss
```diff
@@ -0,0 +1,36 @@
+.l-cluster {
```
alexkilgour, line 1: Is there a reason for using `l-`? Our convention for components is `c-`

sturobson: There was some [previous discussion around this](https://github.com/springernature/frontend-toolkits/pull/577#issuecomment-927993134).

I guess something like the front-end playbook would need an update to classify this more?

Unlike component things like a button, or a header - the `l-` layout primitives would be something added to the `c-` component to avoid over duplication of CSS (or it could used on its own). And like margin, or padding, it's not really a `u-` utility as they're, generally, doing more than one thing in the CSS.

There is now precedent with `l-stack`, `l-with-sidebar`, and now `l-cluster` in the front-end toolkits. But maybe a retrospective front-end open space ticket might be useful?

These layouts are based on @Heydon and [Andys](http://twitter.com/piccalilli_) [Every Layout](https://every-layout.dev).

Heydon: @alexkilgour Yep, basically these don't represent lexical components, more like lower-level helpers. Something between utilities and components. Hope that makes sense.

Heydon:
@alexkilgour @benclark I have updated the package to use Alex's new dynamic partials mechanism to display multiple horizontal alignments from the one `/view` template. 

https://user-images.githubusercontent.com/683558/144480271-70d1807a-7c81-45dd-9f26-faa2825d1636.mp4

Is this okay to be merged now?

**Side note:** I have named the template in `/view` after the package. In the springer.com Proof Of Concept we dynamically import templates and this convention feels right to me? Of course, we import from the demo folder for now, since view folders were scarce. Each file is named `index.hbs` in `/demo` which we rename after the package on the fly. I'm not sure quite what the best convention is here but I note the **springer-media** `/view` template is called `media.hbs`. This isn't a problem where there is only one template in `/view` because we can just take whatever is there and rename as with the `index.hbs` templates. But it would become difficult if there were multiple templates in `/view`. Perhaps something to avoid?

I have written an executive summary of the springer.com POC here: https://docs.google.com/document/d/19D85MT848x4441zpDsZXULlyWo6uCcvV37uQzUsTucs/edit#

benjclark (approved):
I agree that naming/prefixing the templates in view with the name of the component is a good convention for us to go with. It is also in-line with how we name Sass variables in 10-settings for the component.

PR #624, opened 2021-11-26
Use new images in demo functionality

alexkilgour:
- update the `frontend-package-manager` to use the new images in demo functionality from `util-package-renderer`

### global-corporate-footer, springer-header, springer-listing, springer-publisher-footer
- update demo to handle images

### springer-media
- add sample image to the demo, and render that in the distribution file
- move consumable template to `view` folder
- Add option for the image on the right
- Improve demo to show different examples
- Remove unnecessary call to default `brand-context` in demo
- No longer reliant on `u-text-lg`


toolkits/springer/packages/springer-media/demo/main.scss
```diff
@@ -2,8 +2,6 @@
 
 // we must not add `node_modules/` to the Sass includePaths config,
 //   so use relative paths here. Paths should be relative to this file.
-@import '../node_modules/@springernature/brand-context/default/scss/core.scss';
-@import '../node_modules/@springernature/brand-context/default/scss/enhanced.scss';
```
alexkilgour, line 6: each of the imprint brands (nature, springer, springernature) imports default the builds on it, so no need to import it here

toolkits/springer/packages/springer-media/scss/50-components/_media.scss
```diff
@@ -67,6 +67,7 @@
 
 .c-media__title {
 	margin-bottom: $media--title-bottom-spacing;
+	font-size: $context--font-size-lg;
```
alexkilgour, line 70: In the readme, the html example specified using `u-text-lg`. We don't want to depend on utility classes in components as they should be self-contained. We are able to specify this within the component CSS without using the utility class. Shown here.

The utility class was also not used in the demo, which means the final rendered component looks different (and may well look different in production if the utility class was not added). See example below

example

nickcall (approved):

PR #625, opened 2021-11-29
the one that resizes all the fonts and turns them all to rems. 

sturobson:
As discussed during an [front-end open space](https://github.com/springernature/frontend-open-space/issues/293) [slack conversation](https://springernature.slack.com/archives/C1HGRAQ8G/p1632402667047500).

This PR should normalise things so that the bast font size is set back to `100%` and all font-sizes are set with the `rem` unit accordingly. 

This is a pre-cursor to adding Design Tokens to the front-end toolkits as a way to create and generate colours, spacing, z-index, media queries, font-sizes etc etc.

Hopefully - I've been diligent enough to make sure that nothing visually changes across the brands and components that I've made edits to. 

What I have done is change the font-size on the `html` element from `62.5%` to `100%`. I have then changed the font-size on the `body` from `1.8rem` to `1.125rem`. This _should_ mean any values set in `em` will be the same as before. And I've changed the font-size for various text elements from `px` to `rem` or changed the `rem` value to match the `html` value.

I've put this as a Draft PR - so that it can be tested.


sturobson:
I've not altered these two `rem`/`em` values as I'm not sure what they're based on as a root font size - https://github.com/springernature/frontend-toolkits/blob/master/context/brand-context/nature/scss/10-settings/typography.scss#L40

sturobson:
- [ ] I've not changed any `margin`/`padding` values yet

ricky-lau:
> I've not altered these two `rem`/`em` values as I'm not sure what they're based on as a root font size - https://github.com/springernature/frontend-toolkits/blob/master/context/brand-context/nature/scss/10-settings/typography.scss#L40

There are examples of it being used on headings and article titles on [nature homepage](https://nature.com). 

`-0.01875rem` is used from `2.0rem` and `2.4rem` font-sizes.
`-0.0625rem` is used from `3.2rem` font-size.

context/brand-context/nature/scss/10-settings/typography.scss
```diff
@@ -8,37 +8,37 @@ $context--font-family-serif: if($is-nature-branded, (Harding, Palatino, serif),
...
+$context--font-size-h4-min: 1rem; // Body typography size
 
 $context--line-height-h1-min: $context--line-height-h2;
 $context--line-height-h2-min: $context--line-height-h3;
 $context--line-height-h3-min: $context--line-height-h4;
 $context--line-height-h4-min: $context--line-height-h4;
 
-$context--letter-spacing-min: -0.01875rem;
-$context--letter-spacing: -0.0625rem;
+$context--letter-spacing-min: -0.01875rem; // TODO - What is this?
```
benjclark, line 40: These variables are used by https://github.com/springernature/frontend-toolkits/blob/master/context/brand-context/nature/scss/30-mixins/typography.scss for setting letter-spacing on heading elements. We should probably update the values to be relative to 100% on the html.

sturobson: > We should probably update the values to be relative to 100% on the html.

I can do this as part of the other changes - so hopefully, visually, _nothing_ changes<sup>*<sup>

<sup>*</sup> famous last words.

benjclark: Oh are those changes happening in a different PR?

I think it's just a case of updating to:
```
$context--letter-spacing-min: 0.001171875rem;
$context--letter-spacing: -0.0390625rem;
```

sturobson: no, they're as part of this PR - I need to change the title a little I guess and add more detail. 

I've gone through and changed any css rules that effect type that is set in `px` or `rem` to match a root font-size of `100%`.
If there are rules set in `em` I have left them as they should respect the body font size that is `18px` (or `1.125rem` now (it was `1.8rem`).

So here, because they're set in `rem` they need changing. 

I got

```
$context--letter-spacing-min: -0.011715625rem; // in Safari this is 0.18745000660419464px not 0.1875px
$context--letter-spacing: -0.0390625rem;
```

So will double check and update accordingly.

benjclark (approved):
Nice one

Heydon:
@benjclark Stu is on it, I relayed what I've been doing in the packages 😅

sturobson:
> just realised there's no HISTORY.md or package.json version number updated included in this PR?

@benjclark I've just gone through and added relevant text to the `history.md` files and updated the package version numbers.

I've also updated the `brandContext` version in the package.json files for each to match the new version number, I'm assuming that is correct?

sturobson:
I'll now fix those merge conflicts

context/brand-context/HISTORY.md
```diff
@@ -1,7 +1,12 @@
 # History
 
+## 18.0.0 (2021-12-03)
```
benjclark, line 3: If it's a new major version can you include `BREAKING:` before the bullets (see version 11 in this file)

benjclark: Although thinking about this more, this is an interesting one. Because in theory nothing should be broken after these changes. Which means this could be a minor version increment instead.

benjclark: Or did you intentionally bump major because there is potential for something to be broken but we are unsure? If that's the case maybe it's safest to declare it as breaking. 🤔

sturobson: > Breaking major change seems safest.

As alluded to in my comment regarding project specific CSS - let's go with breaking. 👍🏼

benjclark (dismissed):
https://user-images.githubusercontent.com/5796370/144621640-4a565fbe-a4e6-40f4-bfa1-85f190b22e96.mp4

benjclark (approved):

PR #626, opened 2021-11-30
Add keyframes to springer brand context

nickcall:
Adding keyframes to the springer brand it can then be removed from multiple scss files it sits within across oscar article stylesheets


alexkilgour (approved):
beautiful

PR #627, opened 2021-12-01
Add yellow to the context colours for use in the keyframes

nickcall:
Adds the colour `yellow` to the context-colours so that it can be used by the keyframe settings


axemonkey (approved):
LGTM 👍

PR #628, opened 2021-12-02
Fix page elements bleeding through springernature-modal

aliuk2012:
I came across an issue where I had some elements on the page that had
an z-index and these were bleeding through the modal. To overcome this,
I added a z-index to the modal when its in open state and increased the
index to a stupidly high number so its less likely other elements on
the page may ever bleed through again.

image


nickcall (approved):
LGTM

nickcall (approved):

PR #631, opened 2021-12-07
the one that corrects a few mis-calculated font sizes and catches some components that missed the update.

sturobson:
In a recent conversation about nature.com all of a sudden having large fonts I noticed that I had made a few calculative errors in converting the font sizing from `px` to `rem` and changing the `rem` value now the root/html is set to `100%`.

This PR fixes that.

I guess I need to create a "patch" release for the components and the context? 

The issue with nature.com relates to the Oscar repo and its ~120 instances of `font-size` that was relying on the root/html being `62.5%`. Nick from the Orion team is making an update to fix that.

todo:

- [x] update `package.json` version number for affected npm pacakges
- [x] update `history.md` to comment on changes

--- 

During the updates here, and the work being carried out to update Oscar it seems a few components missed the memo and didn't get correct font-size updates. I've also made those changes here.

*Note:* There is one package I did not update - [the Springer Nature submission header] looks like the it is already expecting the root/html font size to be 100%. Because of this I have made no edits.


benjclark (dismissed):

toolkits/nature/packages/nature-footer/package.json
```diff
@@ -1,9 +1,9 @@
 {
     "name": "@springernature/nature-footer",
-    "version": "1.0.0",
+    "version": "2.0.0",
     "license": "MIT",
     "description": "Nature branded footer",
     "keywords": [],
     "author": "Springer Nature",
-    "brandContext": "^9.0.1"
+    "brandContext": "^17.0.0"
```
alexkilgour, line 8: should this be the latest `18.0.1`?

toolkits/nature/packages/nature-header/package.json
```diff
@@ -1,11 +1,11 @@
 {
   "name": "@springernature/nature-header",
-  "version": "5.2.3",
+  "version": "6.0.0",
   "license": "MIT",
   "description": "Publisher level header",
   "keywords": [],
   "author": "Springer Nature",
-  "brandContext": "^11.1.0",
+  "brandContext": "^17.0.0",
```
alexkilgour, line 8: should this be the latest `18.0.1`?

toolkits/nature/packages/nature-hero/package.json
```diff
@@ -1,9 +1,9 @@
 {
     "name": "@springernature/nature-hero",
-    "version": "1.1.1",
+    "version": "2.0.0",
     "license": "MIT",
     "description": "Nature branded hero",
     "keywords": [],
     "author": "Springer Nature",
-    "brandContext": "^11.1.1"
+    "brandContext": "^17.0.0"
```
alexkilgour, line 8: should this be the latest `18.0.1`?

toolkits/nature/packages/nature-sort-by/package.json
```diff
@@ -1,11 +1,11 @@
 {
   "name": "@springernature/nature-sort-by",
-  "version": "1.1.0",
+  "version": "2.0.0",
   "license": "MIT",
   "description": "Nature sort by ui dropdown",
   "keywords": [],
   "author": "Springer Nature",
-  "brandContext": "^11.1.0",
+  "brandContext": "^17.0.0",
```
alexkilgour, line 8: should this be the latest `18.0.1`?

toolkits/springer/packages/springer-listing/package.json
```diff
@@ -1,9 +1,9 @@
 {
   "name": "@springernature/springer-listing",
-  "version": "3.1.1",
+  "version": "4.0.0",
   "license": "MIT",
   "description": "Display of scannable content typically within a list",
   "keywords": [],
   "author": "Springer Nature",
-  "brandContext": "^9.0.3"
+  "brandContext": "^17.0.0"
```
alexkilgour, line 8: should this be the latest `18.0.1`?

toolkits/springernature/packages/springernature-header/package.json
```diff
@@ -1,9 +1,9 @@
 {
   "name": "@springernature/springernature-header",
-  "version": "2.2.1",
+  "version": "3.0.0",
   "license": "MIT",
   "description": "Springer Nature branded header component",
   "keywords": [],
   "author": "Springer Nature",
-  "brandContext": "^10.0.0"
+  "brandContext": "^17.0.0"
```
alexkilgour, line 8: should this be the latest `18.0.1`?

toolkits/nature/packages/nature-footer/package.json
sturobson: I found an issue with Travis yesterday when I set everything to `^18.0.0` so dropped it back to `^17.0.0` to make Travis happy.

@benjclark suggested a doing quick 'post merge' PR to move this to `^18.0.0` for the components, which I didn't get to because we found the font size issues with Oscar.

I _think_ the `^` means it should look for `v18.0.0` so perhaps we merge this in now and I go through and make the changes to `18.0.1` on the components after the brand context os updated to npm?

alexkilgour: Oh of course yes, because the v18 won't be published at the point that travis checks if it exists for the other components. Would these changes not break the font sizing if used with 17?

sturobson: hmmm, I guess they might do. 

As `18.0.0` is now live on NPM I could change any instances of `^17.0.0` I have made to `^18.0.0` and that'll clear it up?

alexkilgour (approved):

toolkits/nature/packages/nature-footer/package.json
alexkilgour: Yep let's go with that

sturobson:
As mentioned in a Slack conversation.

There's a variable `$base-font-size` that's used as part of a Sass function to create a `rem` based value. 

This was set at `10` which worked when the root/html font size was `62.5%` but as the front-end toolkits now use a root/html font size of `100%` the value needs to change to `16`.

I've made that change in this PR too. 

Ultimately, however, the function shouldn't really exist, even more so when Design Tokens finally (finally!) get implemented as we can be more explicit in the use of computed Sass variables and the naming of them.

benjclark (approved):

PR #632, opened 2021-12-07
new springer-hero component

Heydon:
* Following in Nature’s footsteps
* Intrinsically responsive
* For the new springer.com
* Aesthetic-wise it may well be iterated somewhat

https://user-images.githubusercontent.com/683558/145028021-c8bff96e-4059-4cf1-b7d1-2830f34f8c15.mp4


sturobson:
I'm wondering if this could/should be using the `l-sidebar` as well to help the layout changes?

toolkits/springer/packages/springer-hero/scss/50-components/springer-hero.scss
```diff
@@ -0,0 +1,42 @@
+.c-hero {
+	display: flex;
+	flex-wrap: wrap;
+	background-color: $hero--background-color;
+	color: #fff;
```
sturobson, line 5: The default brand context has `$context--foreground-light` as `#fff` but that gets overridden if used here with tbe Springer brand variable.

Just making a note - I think the `shared` colours for the default brand context could do with variables for black and white (and a greyscale) so we don't have to write hex codes.

toolkits/springer/packages/springer-hero/view/springer-hero.hbs
```diff
@@ -0,0 +1,19 @@
+<div class="c-hero">
+	{{#if imageLeft}}
+		<div class="c-hero__image">
+			<img src="{{imageSrc}}" alt=""/>
```
sturobson, line 4: Would it be wise to add a variable for the `alt=""` too?

toolkits/springer/packages/springer-hero/scss/10-settings/springer-hero.scss
```diff
@@ -0,0 +1,3 @@
+$hero--padding: spacing(24);
```
sturobson, line 1: you don't seem to be using this? Should it be used on line 32 of the springer-hero.scss file (and the value changed accordingly).

toolkits/springer/packages/springer-hero/scss/50-components/springer-hero.scss
```diff
@@ -0,0 +1,42 @@
...
+.c-hero__image img {
+	position: absolute;
+	width: 100%;
+	height: 100%;
+	object-fit: cover;
+}
+
+.c-hero__content {
+	flex-basis: 0;
+	flex-grow: 666;
```
sturobson, line 24: > for it a human number

Heydon: 😇

toolkits/springer/packages/springer-hero/scss/10-settings/springer-hero.scss
Heydon: Well spotted.

toolkits/springer/packages/springer-hero/view/springer-hero.hbs
Heydon: I am going to consider it decorative (since it just illustrates the title, like "Astronomy"). Ideally, it _would_ have `alt` text anyway, but I'm worried about people using poor/placeholder `alt` text, which is worse than nothing. I see this a lot on the current site

Heydon:
> I'm wondering if this could/should be using the l-sidebar as well to help the layout changes?

I thought about this, but it would introduce a cross-dependency? I think we're avoiding those wherever possible.

sturobson:
> I thought about this, but it would introduce a cross-dependency? I think we're avoiding those wherever possible.

True, and the front-end playbook does say not to do this so I guess it shouldn't be introduced. Although, one could suggest the brand-context is a cross-dependency. 

LTGM

sturobson (approved):
Aw yeah!

PR #634, opened 2021-12-09
Global details component (formerly springer-expander)

Heydon:
* A robust expander component
* Optional heading and/or list semantics
* Uses `<details>` so JS-independent
* Enhanced focus styles where `text-decoration-color` is supported

https://user-images.githubusercontent.com/683558/145385851-106dec11-449c-4b1e-a197-1b3c76521b1a.mp4


morgaan:
Hi Heydon :wave:, I’ve very pleased to see this addition to the toolkit 👏 🙏

Were you aware though of the existence of this one component https://github.com/springernature/frontend-toolkits/blob/master/toolkits/global/packages/global-expander/README.md

What about `springer-accordion` as in https://www.w3.org/TR/wai-aria-practices-1.1/#accordion?

I actually wonder if just as the `global-expander` coud be promoted straight to `global-accordion`. We'll for sure soonish need it on SpringerNature brand, and I'm not sure these kind of comp do need strong branding.

Heydon:
@morgaan Hi! This was originally an interface on top of global-expander! We actually use global-expander already for some other stuff on the WIP springer.com 🙂

However, @sturobson suggested the use of details/summary since it doesn't have the JavaScript dependency. But it has other advantages. For example, you can easily apply the `open` property to have the component open by default (on page load).

A set of consecutive expanders would be similar to an accordion except it would lack a couple of accordion features I don't actually think are beneficial:

* Arrow keyboard navigation: in my experience, very few keyboard users navigate by arrow key or know where to expect it being supported. There's also issues with overriding the arrow keys' default behavior, especially in screen readers
* When you open one expander it closes any others that are currently open: this can be disorientating and cause weird scrolling/viewport behavior. 

There's some good info in the ARIA practices guide but it's not a spec' and you don't have to follow the patterns exactly (at least in many cases) to satisfy WCAG requirements. Generally, I think simpler is better. 

Thanks for taking a look at this!

aliuk2012:
To me it looks like an accordion.  "Expander" is also very ambiguous, eg could I use it to "expand/collapse" one or more element on page (i.e like a show/hide type feature), looking at the source code the answer is no.  What about "springer-details". I could see this being useful for other brands, could we make a generic version of it?

For what its worth this is what GDS design system have for these similar components
https://design-system.service.gov.uk/components/details/
https://design-system.service.gov.uk/components/accordion/

Heydon:
> To me it looks like an accordion. "Expander" is also very ambiguous, eg could I use it to "expand/collapse" one or more element on page (i.e like a show/hide type feature)

I've explained my issues with accordions to @morgaan. I'm a bit confused it's not clear in the code that it expands/collapses content, since the README describes this behavior and the demo demonstrates it?

I have considered springer-details, that's a good idea I think! +1 I wonder if others would support that name change.

sturobson:
details was the component name we went with when creating something similar (using `<details>` and `<summary>` for the Visual Framework when I was at EMBL. 

As expander is used in the global components and there looks to be positive vibes to make this a global component in the (near) future. A change now to details would help that move. 

Unless we go 'all out' now and get it in the global components with some changes in this PR.

morgaan:
@Heydon Thank you for the clarification and I must admit I had very little time between 2 meetings and the notification in the frontend-pr channel raised my attention because, first off it is a great components to have and will need it soonish, and second the name got my attention as we are using the global-expander already...

I'm all in for a `global-details` (yes global would be amazing 😊).

Also thank you for the advice on the Authoring Practice guide that is a resource I may value to an extreme extends it does not need.

jon-stevens:
Looks great!

jon-stevens (approved):

morgaan:
On Researcher Portal which is under Springer Nature brand, we have a story most likely to be tackle early next year that would definitely use this component, if this aid weight in the balance toward making it a global component 😛

Heydon:
> On Researcher Portal which is under Springer Nature brand, we have a story most likely to be tackle early next year that would definitely use this component, if this aid weight in the balance toward making it a global component 😛

Sure, I'm happy to help there. @jon-stevens thanks for the approval but I think I will rename and promote as discussed. Perhaps I could request a further quick review when I've sorted that 😬

aliuk2012:
> @aliuk2012
> 
> > To me it looks like an accordion. "Expander" is also very ambiguous, eg could I use it to "expand/collapse" one or more element on page (i.e like a show/hide type feature)
> 
> I've explained my issues with accordions to @morgaan. I'm a bit confused it's not clear in the code that it expands/collapses content, since the README describes this behavior and the demo demonstrates it?
> 
> I have considered springer-details, that's a good idea I think! +1 I wonder if others would support that name change.

@Heydon, 

Sorry I didn't mean to imply that the README or demo weren't clear enough. I was just trying to say that when given a list of components (eg just visiting https://github.com/springernature/frontend-toolkits/tree/master/toolkits/global/packages) it would be difficult to spot which component would be suitable for a particular feature that a dev might be looking for without going into each one to see what they are for and what they do (obviously some sort of preview app would make it easier which I know there is something in the pipeline). Naming things is difficult 😄 👍 

Looks like the everyone agrees it should be a `global-details`. Let me know if I can help in any way to get this over the line.

🍻 🍻 🍻 🍻 🍻 🍻 🍻

benjclark:
**Global Expander**
This has been talked about recently and the consensus seemed to be that it is a javascript helper, not a component. So I think sometime in the near future Global Expander and any other javascript helper such as those helpers found in Global Javascript will get deprecated and moved in brand context instead.

**Finding a component for your needs**
Hopefully the Elements site that is still work in progress will help significantly with developers finding the right component for their needs without having to search through code in such a time consuming way.

Heydon:
Okay, I've promoted the component to the global level and added simple settings files for each brand . Screen shots showing focus styles:

## Default

default

## Nature

nature

## SpringerNature

springernature

## Springer

springer

> Looks like the everyone agrees it should be a global-details. Let me know if I can help in any way to get this over the line.

Thanks, @aliuk2012! There's only so much context/explanation you can put in a name alone, of course. Arguably, "details" is even less descriptive (and relies on an awareness of the HTML element 😅) but I like that this disambiguates things regarding the `global-expander` JavaScript.

This is ready for another check! 🙏

sturobson:
Looking p. fine to me. 

man holds hand to mouth, the word "damn" scrolls across the image

Heydon:
Anyone have an idea what's happening here? 

image

sturobson:
I don't know why it's complaining as that was updated today - https://github.com/springernature/frontend-toolkits/blob/master/toolkits/global/packages/global-facet/HISTORY.md#101-2021-12-10

benjclark:
Looks to me like some of the version updates I made to packages in my PR earlier today have not made it to npm. Specifically seems to affect `global-skip-link` and `springernature-header`. The Travis build for my PR confirms the new versions were supposed to get published to npm - however for some reason they have not yet. This might be a delay on npm's side which I have seen happen before.

benjclark:
image

sturobson:
> image

@benjclark npm seems to be all fine and dandy again now.

benjclark (approved):

PR #636, opened 2021-12-09
the one that updates version of global-meta brand context version required

sturobson:
We move the version of the `global-meta` component to use an older version of the `brandContext` so that it we could get Travis to pass the tests.

Now we've done that, this PR just updates the version of the Brand Context that should be used to the latest - `18.1.1`

### DIFF
--- a/toolkits/global/packages/global-meta/package.json
+++ b/toolkits/global/packages/global-meta/package.json
@@ -1,9 +1,9 @@
 {
   "name": "@springernature/global-meta",
-  "version": "6.0.2",
+  "version": "6.0.1",
   "license": "MIT",
   "description": "Brief information about the content",
   "keywords": [],
   "author": "Springer Nature",
-  "brandContext": "17.0.0"
+  "brandContext": "^18.1.1"
 }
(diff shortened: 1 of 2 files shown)


toolkits/global/packages/global-meta/package.json
```diff
@@ -1,9 +1,9 @@
 {
   "name": "@springernature/global-meta",
-  "version": "6.0.2",
+  "version": "6.0.1",
```
benjclark, line 3: I think we will need to go with `6.0.3` here as `6.0.2` has been published to npm already so any consumer with `global-meta: "^6.0.0"` in their package.json would get `6.0.2` unless we make this `6.0.3`

sturobson: D'oh!

benjclark:
```suggestion
  "version": "6.0.3",
```

sturobson: D'oh, again. I updated the HISTORY.md, not the version.

benjclark (approved):

PR #640, opened 2021-12-10
the final? fix for font-sizing

sturobson:
Fixes an `em` value that still thinks the root font-size is `62.5%` to a `rem` unit knowing the root font-size is `100%`.

I was so preoccupied with wether or not I had changed all the `rem` values, I didn't stop to think about `em`s. 

Ian Malcom, from Jurassic Park, gesticulating and saying 'well, there it is.'


Heydon (approved):
LGTM

PR #643, opened 2021-12-16
 Add nature settings file for global-lists

ricky-lau:
add nature override on description list


toolkits/global/packages/global-lists/HISTORY.md
```diff
@@ -1,4 +1,8 @@
 # History
+
+## 5.1.0 (2021-12-16)
+    * Add nature settings file
```
benjclark, line 4: I think it is probably worth calling out the override and why here

benjclark (approved):

PR #645, opened 2021-12-22
Springer cards

Heydon:
Making cards more flexible to allow for irregular image shapes (portraits/avatars). From HISTORY.md:

## 7.1.0 (2021-12-29)
    * Option to remove faux block link style

## 7.0.0 (2021-12-22)
    * BREAKING:
        * Template moved to `/view`
        * List semantics by default
        * Shared `level` (`aria-level`) value
    * Support for portrait shaped images (avatars)
    * `noShape` option to remove border, background, and box-shadow
    
## Portrait handling, including landscape "double" portraits

set_of_4


toolkits/global/packages/global-card/HISTORY.md
```diff
@@ -1,5 +1,16 @@
 # History
 
+## 7.1.0 (2021-12-29)
```
benjclark, line 3: I think all of these changes should be lumped in to v7.0.0 because otherwise there is no way a user could install v7.0.0 if they tried to because it was never published.

toolkits/global/packages/global-card/demo/context.json
```diff
@@ -1,11 +1,80 @@
 {
-	"image": {
-		"src": "https://picsum.photos/500/500",
-		"alt": "descriptive alternative text"
+	"cardsDefault": {
```
benjclark, line 2: In this demo json config file could we portray some data that defines some of the variables used in the hbs template i.e. `noList`, `noListItem`, `level` and `aspectRatio`. I know they are mentioned in the README and in the hbs template but I think it will help consumers understand the implementation better if they can see example data that implements those. Particularly in understanding and tracing `{{>card parent=..}}` in the demo hbs

Heydon: @benjclark Good idea! Thanks. I'll do separate demos for each.

toolkits/global/packages/global-card/HISTORY.md
Heydon: @benjclark Good point!

Heydon:
@benjclark I added more demos but have left out `noList`/`noListItem` since it has no visual effect and `noShape` because this is built into `portrait` anyway. What do you think?

benjclark:
Ok cool, what about `level` ?

Heydon:
@benjclark Also no visual effect? I'm not sure how important it is the differences are visual, but it might be confusing where they aren't. I'm on the fence.

benjclark:
@Heydon I like the idea of `level` being included in one of the demo examples because it while it has no visual effect it does help show the developer where that data should be defined in the data model at a quick glance instead of having to work it out in their head from:
```
<h3 class="c-card__title"{{#if ../parent.level}} aria-level="{{../parent.level}}"{{/if}}>
```

And in addition, I like the idea of providing an example that changes aria-level as it helps to ensure consumers think about that a11y consideration. Lowers the chance of them overlooking it if we give a nice example demonstrating it.

Tell me if I'm being too pedantic 😄

Heydon:
@benjclark I improved the docs with some more example JSON and added that level demo. Good to go? Thanks for the pointers/ideas.

benjclark (approved):
thank you!

PR #646, opened 2021-12-29
Adds Nature Section Heading component to replace deprecated Nature He…

benjclark:
…adings component

This component replaces the soon to be deprecated Nature Headings component. The README has been written in accordance with the discussions between Amy Hupe, James Alegria and Ben Clark.

The component has also been updated to include `__icon` and `__link` classnames so that the implementation does not depend on the use of utility classnames.

PR #647, opened 2021-12-29
Removes Nature Headings component and updates DEPRECATED.md

benjclark:
This removes Nature Headings component as it is being deprecated and replaced by a new component [Nature Section Heading](https://github.com/springernature/frontend-toolkits/pull/646).

**Reasoning**

Nature Headings was created to act as a bucket to hold any classnames for different visual designs for section headings on Nature. This is similar to [Global Lists](https://github.com/springernature/frontend-toolkits/tree/master/toolkits/global/packages/global-lists/scss/50-components) where there are different list classnames and therefore filenames, all stored in the same component. However, we no longer want to implement components in this way; going forward components should be single ui features. And in addition to that, there is only one classname in Nature Headings; c-slice-heading, so the whole idea of it being a home for a variety of different visual designs is currently pointless.

Therefore we are recreating c-slice-heading as the more appropriately named c-section-heading component. The new component is a single ui feature.

After this change has been made I will run:
```
npm deprecate @springernature/nature-headings https://github.com/springernature/frontend-toolkits/blob/master/DEPRECATED.md
```

which will ensure anyone installing Nature Headings will get a deprecation warning. It will also mean the component stops showing up in npm searches.


sturobson:
As a 'double feature' I wonder if adding something to the Sass file using `@warn` would be useful too. 

I for one rarely look at npm deprecation warnings because often it's waiting on X which is waiting on Y which is waiting on Z.

Adding a 'one liner' to a Sass file would, I feel, be more of a sticking point as the notice is printed out in the terminal every time Sass compiles. 

Something like (because I guess this should be discussed, standardised, and agreed (if you think it's worthwhile)) –

`@warn 'This component has been deprecated since vX.X.X. Please use name-of-component instead';`

alexkilgour:
> As a 'double feature' I wonder if adding something to the Sass file using `@warn` would be useful too.
> 
> I for one rarely look at npm deprecation warnings because often it's waiting on X which is waiting on Y which is waiting on Z.
> 
> Adding a 'one liner' to a Sass file would, I feel, be more of a sticking point as the notice is printed out in the terminal every time Sass compiles.
> 
> Something like (because I guess this should be discussed, standardised, and agreed (if you think it's worthwhile)) –
> 
> ```css-scss
> @warn 'This component has been deprecated since vX.X.X. Please use name-of-component instead';
> ```

I like this idea. Let's get this PR through and I'll look at getting an update across all deprecated packages

alexkilgour (approved):
This looks good 👌

benjclark:
@alexkilgour So I tried to run 
```
npm deprecate @springernature/nature-headings https://github.com/springernature/frontend-toolkits/blob/master/DEPRECATED.md
```
but I get:
```
npm ERR! code E404
npm ERR! 404 Not Found - PUT https://registry.npmjs.org/@springernature%2fnature-headings - Not found
npm ERR! 404 
npm ERR! 404  '@springernature/nature-headings@latest' is not in the npm registry.
npm ERR! 404 You should bug the author to publish it (or use the name yourself!)
npm ERR! 404 
npm ERR! 404 Note that you can also install from a
npm ERR! 404 tarball, folder, http url, or git url.
```

If I run 
```
npm i @springernature/nature-headings
``` 
or run
```npm search @springernature/nature-headings```
it finds the package.

I have also tried:
```
npm deprecate @springernature/nature-headings@0.1.0 https://github.com/springernature/frontend-toolkits/blob/master/DEPRECATED.md
```
and
```
npm deprecate @springernature/nature-headings@0.x https://github.com/springernature/frontend-toolkits/blob/master/DEPRECATED.md
```
but they have the same result. 🤔 

My gut feeling is that it's something to do with the package only ever being published as 0.1.0 (i.e. it never reached 1.0.0). What are your thoughts?

**Comment by benjclark** [comment]
Thanks Stu!

I tried that but unfortunately no luck.

```
npm deprecate @springernature/nature-headings "Nature Headings has been recreated as nature-section-heading (see https://github.com/springernature/frontend-toolkits/tree/master/toolkits/nature/packages/nature-section-heading). For reasoning see https://github.com/springernature/frontend-toolkits/pull/647."
```

produces:
```
npm ERR! code E404
npm ERR! 404 Not Found - PUT https://registry.npmjs.org/@springernature%2fnature-headings - Not found
npm ERR! 404 
npm ERR! 404  '@springernature/nature-headings@latest' is not in the npm registry.
npm ERR! 404 You should bug the author to publish it (or use the name yourself!)
npm ERR! 404 
npm ERR! 404 Note that you can also install from a
npm ERR! 404 tarball, folder, http url, or git url.
```

PR #648, opened 2021-12-31
Listing reorganization

Heydon:
From HISTORY.md:

```
## 5.0.0 (2021-12-31)
    * BREAKING:
        * Now renders as a list, looping over individual listings
    * New `/view` template 
    * Separate demos for different settings
```

Also uses a new algorithm for wrapping / responsive design.

## Illustration

listing with options


toolkits/springer/packages/springer-listing/scss/50-components/_listing.scss
```diff
@@ -1,80 +1,96 @@
-.c-listing {
+.c-listings {
```
benjclark, line 1: I think this needs to be `c-listing` in order to adhere to normal BEM convention. Or alternatively everything else needs to be `c-listings__`. As the BEM elements and modifiers are extensions of the block, so they should all share the same block value so people can easily see the relationship. What do you think?

toolkits/springer/packages/springer-listing/scss/10-settings/_listing.scss
```diff
@@ -1,17 +1,8 @@
-$listing--font-family: $context--font-family-sans;
-$listing--title-font-family: $context--font-family-serif;
-$listing--title-font-size: 1.125rem;
-$listing--title-font-weight: normal;
-$listing--title-margin-bottom: 4px;
-$listing--title-line-height: 1.3;
-$listing--text-column-flex: 1 0 70%;
-$listing--text-font-size: 1rem;
```
benjclark, line 8: Out of interest what was the thought behind removing font size from the variables and 50-component CSS? On our pages such as https://www.springer.com/journal/4/updates where we use this component we will need to set font-size: 1rem in our application css. Which is totally fine! I just wanted to ask and get onboard with your thinking. I guess setting font-size in this component prevents re-use on other Springer products where the base font-size is different? Makes sense I guess

Heydon: @benjclark Good question. I removed it because I don't see these kinds of basic styles set as variables in any other components. `1rem` should be the default, right? It seemed to me, brand context styles were needlessly creeping into components. But, sorry, I may have misunderstood that. The cards do appear to render in the demos correctly.

toolkits/springer/packages/springer-listing/scss/50-components/_listing.scss
Heydon: @benjclark Ah yes, sorry. I think option 2 is probably best.

sturobson:
The recent changes should mean that:

`html` is set at 100%
`body` is set at the `rem` equivalent of what was there before
the ruleset that had a `font-size` of `13px` on the body, before being set again is removed.

So this is all good tp me -- and with use `rem` it means that with a body `font-size` of `1.125rem` any child element that is set as `1rem` would be `16px` rather than `18px` which would happen if we used the `em` unit.

toolkits/springer/packages/springer-listing/scss/10-settings/_listing.scss
sturobson: A scene from Batman Returns where the Bat Signal lights a study in the new Wayne Manor and Bruce Wayne rises from their chair and looks towards it


The recent changes should mean that:

`html` is set at 100%
`body` is set at the `rem` equivalent of what was there before
the ruleset that had a `font-size` of `13px` on the body, before being set again is removed.

So this is all good tp me -- and with use `rem` it means that with a body `font-size` of `1.125rem` any child element that is set as `1rem` would be `16px` rather than `18px` which would happen if we used the `em` unit.

benjclark: FWIW we might be absolutely fine with two springer.com domain websites being divergent in the font-size they use for this component.

And looking now at the springer.com website I help look after we appear to have two different listing components (fail) that use two different font-sizes (utter fail, ffs). 

So forget about this for now. I don't want to slow this PR down loads by blocking it with this consistency discussion. 

If you could send me a link to a page that has implemented `c-listings` at some point in the future that would be great. Then I could start a discussion with FJ and James Alegria etc and show them all the different listing components and their font-sizes across our Springer domain websites and they can make a call on if they should all have the same font-size or not for a consistent UX and also if they should all be the same component.

toolkits/springer/packages/springer-listing/README.md
```diff
@@ -6,72 +6,100 @@ Optional image shrinks besides the title and text as the viewport becomes smalle
...
-            </ul>                
-        </article>
-    </li>
-</ul>
+# Usage
+
+Import the Sass:
+
+```scss
+@import '../node_modules/@springernature/springer-listing/scss/10-settings/listing';
```
benjclark, line 14: Should the filename now be `listings` instead?

Heydon: @benjclark Oh wait you just mean for the scss file. Yeah. FWIW, I'm not totally comfortable this is listing->listings but I saw no other way to encapsulate it as a list, which should be enforced for accessibility and the group styling.

benjclark: image

Maybe it's ok for this component to stay as `c-listing` ? As even in its singular form (listing) its meaning is an entire list or catalogue or something.

Heydon: @benjclark You genius, I hadn't thought of that! I just reverted it all to singular.

PR #650, opened 2022-01-03
remove padding on lists

Heydon:


sturobson (approved):
pesky user agent stylesheets

PR #654, opened 2022-01-06
Allow blocking of new packages

alexkilgour:
Allow the publishing of new packages to be turned off via a configuration option in the `package-manager.json`. This PR will:
- Update the `frontend-package-manager` version to include the new functionality
- Set the configuration to allow new packages (no change to current functionality)

If the new `blockNewPackages` option is set to `true` then:
- Existing packages can _still_ be updated
- Validation will **fail** if a new package is detected in one of the toolkits
- Validation will **pass** and the package can be committed if the version is set to `0.0.0` meaning the package will not be pushed to NPM
- Packages can **not** be updated from `0.0.0`

(included in this update are some minor changes to the CLI output produced by the `frontend-package-manager` and some minor improvements via refactoring)


benjclark:
So what's the thinking behind allowing people to push new components as v0.0.0?

matteo-gildone (approved):

alexkilgour:
> So what's the thinking behind allowing people to push new components as v0.0.0?

Main reason is simplicity. This way all that has to happen is to check whether a lower version exists on NPM (or no version) and can error at the validation stage. If we want to stop uploading at v0.0.0 as well then would need to start getting into checking git hashes to see if this package existed in the previous commit. Without doing that any already existing packages at version 0.0.0 would error.

Secondly it fits with the way this already works - that you can commit a package at v0.0.0 at it will not publish. the reality is that nobody really uses it.

PR #659, opened 2022-01-11
Allow for styling of fill color

Heydon:
Replaces `fill="#666"` with `fill="currentColor"` on all icons, making icons adopt the local `color` by default and allowing theming. 

Not considered a breaking change since it will match the font `color` contrast rules and we don't have visual regression set up in any case.


sturobson:
Where we've changed `fill="#666"` for `fill="currentcolor"` is the associated `color` property in the ruleset set to `#666` already?

Heydon:
> Where we've changed fill="#666" for fill="currentcolor" is the associated color property in the ruleset set to #666 already?

I'm not sure what you mean, sorry. I looked for `stroke` values to replace too but didn't see any.

Heydon:
> Where we've changed fill="#666" for fill="currentcolor" is the associated color property in the ruleset set to #666 already?

Revisiting this. I think I understand what you mean now. Currently, there are no settings for icon colors. This would just make sure the icons honor the `color` for the context _by default_. To control the icon color (make it different to the surrounding text color) this allows you to apply a `color` to the `u-icon` element.

sturobson:
> This would just make sure the icons honor the color for the context by default.

👍🏼 

> To control the icon color (make it different to the surrounding text color) this allows you to apply a color to the u-icon element.

👍🏼

saramorueta (approved):
LGTM 🆗

PR #662, opened 2022-01-13
Springer search

Heydon:
```
## 7.0.0 (2021-01-12)
    * Optional search form and corresponding menu link
```

Demo:

https://user-images.githubusercontent.com/683558/149372437-5f6915a1-233c-4d33-96aa-737f19a0335a.mp4


toolkits/springer/packages/springer-header/view/springer-header.hbs
```diff
@@ -0,0 +1,66 @@
...
+			<li class="c-header__item">
+				<a href="{{href}}" class="c-header__link c-header__link--nav">{{label}}</a>
+			</li>
+			{{/each}}
+		</ul>
+	</nav>
+	{{/if}}
+</header>
+{{#with search}}
+<form role="search" class="c-header__search-form c-header__search-form--no-js" action="{{action}}" id="search" tabindex="-1">
```
saramorueta, line 52: we normally don't add a class to indicate the default non-js behaviour. We rely on the class `'js` existing on a top-level. https://github.com/springernature/frontend-playbook/blob/main/practices/javascript-styling.md Could you use that instead?

sturobson (commented):
do we need any `appearance: none;` to remove any annoying iOS styling?

saramorueta (approved):
LGTM

toolkits/springer/packages/springer-header/view/springer-header.hbs
Heydon: @saramorueta Sorry, that was part of an earlier experiment. I forgot and left it in. Removed now; it's not needed!

PR #664, opened 2022-01-19
Block new component packages

alexkilgour:
Turn on the option to block the publication of new component packages.
Only updates to existing packages are permitted at this time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,8 @@
 
 > Frontend Component Toolkits for the Elements Design System
 
+**Warning**: Publication of _new_ component packages is currently blocked. Updates to existing components are still permitted.
+
 * [Installation](#installation)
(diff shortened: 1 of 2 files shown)


benjclark (approved):

benjclark:
Should we add anything in the readme to say "you can still add packages as 0.0.0 but they wont be published to npm" ?

alexkilgour:
> Should we add anything in the readme to say "you can still add packages as 0.0.0 but they wont be published to npm" ?
> 
> Although to be honest I'm not sure why anyone would want to do that. Perhaps for the time being people should create new components in their applications and earmark them as to be migrated in to Elements when the block gets turned off again.

Yeah, I don't want to advertise that behaviour! it's just a back up in case we need it. Keeping them in your app is the preference

PR #665, opened 2022-01-19
Adds functionality to highlight invalid email addresses to the user

jon-stevens:
We have done some work to display to our users when there is an invalid email address attributed an author on a submission.

Screenshot 2022-02-01 at 11 24 40


foxintherain:
@jon-stevens Can we please use the icons from the brand context?

jon-stevens:
> @jon-stevens Can we please use the icons from the brand context?

There are no relevant icons in the SN brand context. The nearest one in the default brand context is quite different. Is this the one I should be using?

foxintherain:
> > @jon-stevens Can we please use the icons from the brand context?
> 
> @foxintherain The nearest icon in the default brand context is quite different. Is this the one I should be using?

That's the one :) All our icons have an "outline" style, rather than a solid shape.

jon-stevens:
@foxintherain

Thanks. I've updated the icons. We are still using one however - the abstract icon, which I don't think lives in the brand context yet. If it is all correct, I can add that to the BC separately.

foxintherain:
Thanks @jon-stevens ! Yes we don't have an icon for "abstract" in the brand context, so that can be left alone. One thing to add - the alert icon can also use ['red': #c40606](https://github.com/springernature/frontend-toolkits/blob/master/context/brand-context/springernature/scss/10-settings/colors/default.scss), as this is an error message

foxintherain (approved):

PR #672, opened 2022-01-31
Feature introducing Design Tokens

sturobson:
as discussed 'way back when' here's the first pass of the workflow for creating and integrating design tokens for springer nature products using the front-end toolkits.

This PR does a few things:

- [x] creates design tokens for colours, typography, spacing, breakpoints
- [x] generates a `index.scss` file that imports the compiled Sass files so it's a single import for the toolkits.
- [ ] a decision on how best this becomes a package
- [ ] is included as part of the front-end toolkits
- [ ] has relative tokens for actual use. (phase 2 work)
- [ ] has component tokens for actual use. (phase 3 work)
- [ ] more documentation.


sturobson:
👆 github is being a shit, so I'm closing this branch completely and starting a fresh.

PR #675, opened 2022-02-03
darker box lines

Heydon:
The box was very ill-defined, the borders nearly transparent. Used `greyscale(80)` instead, since there was a precedent elsewhere in the code.


sturobson:
hopefully this gets resolved and made easier soon with declarative Sass variables created from the Design Tokens work 🤞🏼

benjclark (approved):

alexkilgour:
@Heydon Seems there is a fix in here for the demo on `global-details`. There will need to be a new version released for this otherwise the demo on the website will not be updated

PR #677, opened 2022-02-08
Simplification and a11y

Heydon:
```
## 8.0.0
    * BREAKING:
        * Some functionality has been removed since it only addressed a special case on springer.com
        * Parent scope usage removed. Now all settings are added to individual cards. This makes it easier to consume the templates and increases flexibility
        * `href` was not templated correctly (hard-coded)
    * Dark mode now supported in the template/data, with adjustments to color for contrast
```


sturobson:
hopefully we will soon get to a point where Design Tokens will generate most/all of the variables found in `10-settings` 🤞🏼

toolkits/global/packages/global-card/scss/50-components/card.scss
```diff
@@ -94,6 +87,11 @@
 			margin-bottom: 0;
 		}
 	}
+
+	a {
+		// WCAG 1.4.1 Use of Color
+		text-decoration: underline;
```
sturobson, line 93: this feels like something that should be in the CSS further up the stylesheet. I'm assuming it's nuked somewhere?

Is `text-decoration: none;` something we should remove wherever it is near the top of the Sass imports?

I'd have expected that `text-decoration: underline;` be the norm, not the exception.

Heydon: It is indeed removed at a high level (https://github.com/springernature/frontend-toolkits/blob/features/design-tokens/context/brand-context/springer/scss/40-base/basic.scss#L11 for springer)

sturobson: I think we can resolve this for now, to get the merged. 

But I think this sort of a11y thing should be discussed further and implemented accordingly. 

If it looks like a link, and links to a page like a link, it's a link. If it's a node of text that looks the same as every other bit of text you won't know if it's a link unless you hover it (I guess) and we'd want our site visitors to be reading the content, not rushing their mouse cursor over it to find any potential links :sob:

toolkits/global/packages/global-card/scss/10-settings/default.scss
```diff
@@ -1,19 +1,16 @@
...
  */
 
-$card--gutter: 16px;
 $card--border: none;
 $card--box-shadow: none;
 $card--font-family: inherit;
 $card--font-size: 14px;
-$card--spacing: math.div($card--gutter, 2);
-$card--padding: $card--gutter;
+$card--spacing: 8px;
```
sturobson, line 10: as discussed via the google meets - this should be eventually coming predetermined and defined from tokens - so the developer won't need to type it out by hand.

sturobson (approved):
As [mentioned in Slack](https://springernature.slack.com/archives/C0GJK53TQ/p1644415298982769?thread_ts=1644411126.994529&cid=C0GJK53TQ) I'm not suggesting it becomes 'the norm' that you and me review each others PRs into toolkits. 

But, I know this is holding up work on the new SPA stuff, and it's a breaking change so it should be 'opt-in' for any other application/site/product using it.

LGTM.

toolkits/global/packages/global-card/scss/50-components/card.scss
```diff
@@ -103,13 +101,6 @@
 // modifiers
 // ---------
 
-// --flush
```
saramorueta, line 106: sorry to open the debate on a closed PR. I've realised this class is hugely used across nature pages and removing it will impact the spacing. Do you think I can put it back? @Heydon

saramorueta: no worries :)

PR #679, opened 2022-02-09
Add missing functions and mixins to default brand abstracts

aliuk2012:
While implementing SpringerNature brand-context to OASiS I discovered
that some functions and mixins where not being included in the abstract
mean that a developer would have to manually import these functions.


alexkilgour (approved):
good spot 👍

PR #689, opened 2022-02-16
Updates strip unit function and triangle mixin in brand context to us…

benjclark:
…e Dart Sass syntax for math


sturobson (approved):
amazed I missed these :-/

benjclark:
@sturobson easily done. I only found it using the sass-migrator tooling

PR #691, opened 2022-02-17
global-grid: update bourbon-neat v2 to v4

aliuk2012:
v2 causes an issue when used with dart-sass as there is a [syntax error with one of the @warns](https://github.com/thoughtbot/neat/blob/2bde983e693a05edce1acc128ebb278e9058c7dc/core/neat/mixins/_grid-container.scss#L22-L23)

This issue was fixed in [v3](https://github.com/thoughtbot/neat/blob/2bde983e693a05edce1acc128ebb278e9058c7dc/core/neat/mixins/_grid-container.scss#L22-L23).

Seeing as bourbon-neat hasn't been updated for years it seems practical to use the last latest release

### DIFF
--- a/toolkits/global/packages/global-grid/HISTORY.md
+++ b/toolkits/global/packages/global-grid/HISTORY.md
@@ -1,5 +1,6 @@
 # History
-
+## 2.0.0 (2022-02-17)
+    * Update bourbon-neat peerDependency version from min v2 to v4
 ## 1.0.0 (2019-02-05)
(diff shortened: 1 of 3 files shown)


alexkilgour (commented):
This looks good, thanks.
This isn't widely used now as most teams have moved away from CSS grid systems (which explains why it hasn't had much love!). Hoping to be able to retire it at some point if we can get teams migrated off 🤞

alexkilgour (approved):

PR #697, opened 2022-02-23
Adds secondary colour scheme to Global Pill

benjclark:
Demo with default brand context settings:
Screenshot 2022-02-23 at 11 40 40

Demo with nature brand context settings:
Screenshot 2022-02-23 at 11 46 25

image

A secondary colour scheme is useful for things like portraying interactivity. For example on Nature we currently use the primary pill colours for interactive ui elements. Having a secondary colour scheme allows us to use GP on noninteractive ui elements.


saramorueta (approved):
LGTM 👍

PR #698, opened 2022-02-23
Remove screenshots from nature readme intros

amyhupe:
This PR removes component screenshots from the Nature Header and Nature Section Heading READMEs.

We are now able to render component demos at the top of the page on the Elements website, so the screenshots are unnecessary and visually repetitive.


Heydon (approved):
sehr gut!

PR #700, opened 2022-02-24
Adds borderless modifier to Global Pill

benjclark:
Allows us to use bordered pills (helps if something is interactive) and borderless pills


ricky-lau (approved):
💊

PR #701, opened 2022-02-25
Fix focus of autocomplete items for NVDA with Firefox (Windows)

ricky-lau:
* BUG: unable to move focus to select autocomplete items when using NVDA with Firefox (Windows) 
      * add role="listbox" to parent container of autocomplete items
      * add role="option" to each autocomplete item  
  * Change tabIndex value from string to integer


benjclark (approved):
Nice one! 👍

PR #705, opened 2022-03-03
[MC] Fix mistakes in template around `buttonProperties`

morgaan:
- Patch `buttonProps` to `buttonProperties`
- Add extra wrapping curlies to avoid conversion to HTML entities


alexkilgour (approved):
Good catch 😄

PR #707, opened 2022-03-10
Updates nature components to ensure no post-install step

benjclark:
This is needed to remove the post-install step during package publication, as npx commands for `@springernature/util-context-warning` cause issues in CI


ricky-lau (approved):
🪂

PR #711, opened 2022-03-15
springer listing content with line clamp

Heydon:
```
## 5.0.3 (2022-03-15)
    * Support teaser text with `line-clamp` for truncation
```

### DIFF
--- a/toolkits/springer/packages/springer-listing/demo/context.json
+++ b/toolkits/springer/packages/springer-listing/demo/context.json
@@ -14,9 +14,10 @@
 					"Research Article",
 					"2 November 2019"
 				],
+				"content": "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Vestibulum nec massa nunc. In tincidunt dolor ex, ac egestas dui blandit vitae. Quisque volutpat nisi sit amet ullamcorper blandit. Cras feugiat, libero eu venenatis luctus, nibh enim posuere odio, ac hendrerit dolor erat vitae erat. Proin nec lobortis velit. Praesent pretium sapien nisi, finibus rutrum orci pretium id. Suspendisse fringilla sem ut viverra facilisis. Maecenas eu neque mattis, fermentum mi vitae, pretium sem. Cras semper neque id cursus finibus. Donec sed malesuada est. Pellentesque id elit quam. Nullam convallis risus tellus, vitae posuere elit commodo sit amet.",
 				"image": {
-				"imageSrc": "https://picsum.photos/910/650",
-				"imageAlt": "Descriptive alternative text"
+					"imageSrc": "https://picsum.photos/910/650",
+					"imageAlt": "Descriptive alternative text"
 				}
 			},
 			{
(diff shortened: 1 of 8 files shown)


toolkits/springer/packages/springer-listing/scss/50-components/_listing.scss
```diff
@@ -95,3 +94,12 @@
 	margin-right: $listing--view-options-icon-margin-right;
 	fill: currentColor;
 }
+
+@supports (-webkit-line-clamp: 3) {
+	.c-listing__intro {
+		overflow: hidden;
+		display: -webkit-box;
+		-webkit-line-clamp: var(--listing-lines, 2);
```
sturobson, line 102: would we also need `line-clamp` here? For when it gets unprefixed.

(yay! Safari have just (finally) unprefixed `appearance`)

toolkits/springer/packages/springer-listing/scss/50-components/_listing.scss
```diff
@@ -95,3 +94,12 @@
 	margin-right: $listing--view-options-icon-margin-right;
 	fill: currentColor;
 }
+
+@supports (-webkit-line-clamp: 3) {
```
sturobson, line 98: > "truncation is not a content strategy" 

…but progressive enhancement ftw!

toolkits/springer/packages/springer-listing/view/springer-listing.hbs
```diff
@@ -29,6 +29,11 @@
 				{{/each}}
 			</ul>
 			{{/if}}
+			{{#if content}}
```
sturobson, line 32: a tentative little 'um' here - wondering if a generic variable like `content` would conflict with other things? 

I'm a big fan of name-spacing things - so this might just be a 'me thing'.

toolkits/springer/packages/springer-listing/demo/context.json
```diff
@@ -34,14 +35,16 @@
 					"Original Paper",
 					"4 October 2019"
 				],
+				"content": "Morbi luctus eleifend velit ut sodales. Donec semper mauris nec tristique semper. Pellentesque habitant morbi tristique senectus et netus et malesuada fames ac turpis egestas. Morbi mi diam, commodo pharetra pharetra quis, tempus ac felis. Suspendisse at felis ut lorem mollis tristique feugiat eu turpis. Phasellus fringilla lacus ac libero tempus fringilla. Suspendisse sit amet risus dignissim, varius nunc vitae, pellentesque risus. Phasellus tempor dui metus, in mollis sem mollis feugiat. Cras feugiat facilisis quam eu vulputate. Integer ullamcorper arcu in tortor dictum sagittis. Mauris quis leo at nulla bibendum porta non quis ante.",
 				"image": {
-				"imageSrc": "https://picsum.photos/800/700",
-				"imageAlt": "Descriptive alternative text"
+					"imageSrc": "https://picsum.photos/800/700",
```
sturobson, line 40: would we need to remove this little additional whitespace that's found it's way in? Or do we actually need to give it a full tabs width as it's 'inside' the `"image"`?

toolkits/springer/packages/springer-listing/view/springer-listing.hbs
Heydon: Hmm. The `content` name would be localised to the item of data, so I think we're good.

toolkits/springer/packages/springer-listing/demo/context.json
Heydon: Yeah, I need to neaten this up.

toolkits/springer/packages/springer-listing/scss/50-components/_listing.scss
sturobson: I've some thoughts on this which I'll probably mull over this evening. 

I don't think SN has any 'playbook' on custom properties (🔔 potential for an open space conversation) and there's a few ways something like this use case for `line-clamp` could be approached.

I lean to an 'off by default' but that opens up that question of where the custom property could or should be set. Yet, in this same train of thought I'm thinking having it 'on' would make things a little cleaner as I'm sure there are times that the content that's going to be displayed here could be massive.

LGTM … for now.

PR #712, opened 2022-03-21
Fix SpringerNature breakpoints so theres no conflict between Default and SpringerNature components

aliuk2012:
Related to https://github.com/springernature/frontend-open-space/issues/307

This open space issue was discussed 17/03/2022. 

Any thoughts about this change? Is it something we could merge?

### DIFF
--- a/context/brand-context/HISTORY.md
+++ b/context/brand-context/HISTORY.md
@@ -1,4 +1,7 @@
 # History
+## 22.0.0 (2022-05-24)
+    * BREAKING
+      * Updates SpringerNature brand-context breakpoints keys to match default brand-context 
 ## 21.1.1 (2022-05-23)
     * Left aligns basic lists in the springer context
 ## 21.1.0 (2022-05-10)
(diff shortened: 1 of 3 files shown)


context/brand-context/HISTORY.md
```diff
@@ -1,4 +1,7 @@
 # History
+## 21.0.0 (2022-03-21)
```
sturobson, line 2: I wonder if this is more a major breaking change? 

Are we confident that changing things here won't break any existing usage in other project?

Do we need to update any components in the `springernature` toolkits to match this? Or would that be a separate PR?

aliuk2012: Good point about the components in SN toolkits. I checked all of them and none of them use the SN bespoke breakpoints. I think only SN header used breakpoints and it was looking for default branding breakpoints.

I suppose we could have both version of keys inside `$context--breakpoints` as an interim step before completely removing. Either way it should be fairly easy to update legacy key names to the newer default version. 

TBH, the best thing to do is completely remove SN breakpoints settings file altogether and just refer to the default breakpoints like Springer, Nature do.

sturobson: Let's keep both for now but remove the device specific breakpoints. 

I'm happy to go with this. 

If you can tidy up the conflicts I think we can get this merged.

aliuk2012:
@alexkilgour @sturobson I've rebased this branch off master and fixed all the conflicts. Is this something we could merge now?

sturobson (approved):
looks great, thanks for adding the comments to denote what they used to be called 👍🏼

PR #715, opened 2022-03-25
Move shared javascript into the brand-context

alexkilgour:
This PR adds the following to the `brand-context`:
- The contents of the `global-javascript` component
- The contents of the `global-expander` component

Work to do after this PR:
- Deprecate `global-javascript`
- Deprecate `global-expander`

### DIFF
--- a/context/brand-context/default/__tests__/unit/util/onetrust.spec.js
+++ b/context/brand-context/default/__tests__/unit/util/onetrust.spec.js
@@ -1,0 +1,46 @@
+import {checkConsent, isConsentBannerClosed} from "../../../js/helpers/util/onetrust";
+
+describe('checkConsent', () => {
+	const originalDocumentCookie = window.document.cookie;
+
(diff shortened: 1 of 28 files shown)


context/brand-context/default/__tests__/unit/util/onetrust.spec.js
```diff
@@ -0,0 +1,46 @@
+import {checkConsent, isConsentBannerClosed} from "../../../js/helpers/util/onetrust";
```
benjclark, line 1: How many teams are still using OneTrust? Is this still useful?

alexkilgour: I did wonder that, but I wanted the first commit to just be a direct copy of what we have already, if I start altering stuff then deprecate the old components it might get confusing as stuff disappears/changes

benjclark (approved):

PR #716, opened 2022-03-29
remove global-javascript and global-expander

alexkilgour:
Remove `global-javascript` and `global-expander` packages that are now part of `brand-context@20.1.0`


alexkilgour:
NOT NEEDED

PR #719, opened 2022-04-03
optimises SVG icons in default brand context

sturobson:
I've run the SVGs in the `brand-context/default` folder through SVGO to optimise them, keeping the `ID`s, `fill`, `viewbox` and `xmlns` attributes.

Going from a total folder size of 53kb to 27kb. 

This is some preliminary work before pulling icons into the design tokens as a source of truth. 

We could either - merge this (I'll update the package.json and changelog) or I could merge this into the Design Tokens work-in-progress PR and save a double bump in versioning.


sturobson:
deleting this as it's something that could be looked at in the single release move

PR #721, opened 2022-04-04
Card and hero tweaks

Heydon:
```
## 8.2.0 (2022-04-04)
    * Title optional
    * White background-color for springer
    * Body padding removed where no image
```

```
## 0.1.3 (2022-04-04)
    * Image sized to content (and automatically is removed from narrow view)
```

### DIFF
--- a/toolkits/global/packages/global-card/view/global-card.hbs
+++ b/toolkits/global/packages/global-card/view/global-card.hbs
@@ -7,6 +7,7 @@
 	</div>
 	{{/if}}
     <div class="c-card__body">
+		{{#if title}}
 		<h3 class="c-card__title"{{#if level}} aria-level="{{level}}"{{/if}}>
 			{{#if href}}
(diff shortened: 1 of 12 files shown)


toolkits/global/packages/global-card/view/global-card.hbs
```diff
@@ -7,6 +7,7 @@
 	</div>
 	{{/if}}
     <div class="c-card__body">
+		{{#if title}}
```
sturobson, line 10: nice! `{{#if }}` all teh things I say

toolkits/springer/packages/springer-hero/demo/dist/index.html
```diff
@@ -1728,7 +1728,7 @@
 
 .c-hero__image {
   position: relative;
-  min-height: 20rem;
```
sturobson, line 1731: get the content right first, rather than get the CSS to make sure it's the right size ✊🏼

sturobson (approved):
some nice additions to make it more useable imho.

my only thought is - does any of this need adding to the documentation, or do you think the demo is enough? 

approving anyhow :) 

… lgtm

PR #725, opened 2022-04-07
@supports rewrite

Heydon:
I forgot IE11 does not support `@supports` (so negation doesn't work). I have rewritten those rules accordingly.


sturobson (dismissed):
lovely stuff! 

the internet explorer logo with legs and eyes, walking

Heydon:
> lovely stuff!
> 
> the internet explorer logo with legs and eyes, walking

@sturobson Thanks! Now if Travis would just pull its finger out...

PR #731, opened 2022-04-22
Apply h5 styles

Heydon:
* h5 wasn't being applied from the placeholder
* h4/h5 needed to both be sans-serif


sturobson (approved):
this tracks with [the current styles](https://www.sketch.com/s/fa9c2fc9-a179-43f0-b21e-9562c9c17c0c/a/GmRxpP8)

… and as it's a visual bug I think the patch release version number change is 'on point'. 

LGTM

sturobson:
You may merge, when ready.

Grand Moff Train saying 'you may fire when ready'

sturobson:
aside - I think it would be worthwhile to revisit how this typography is created via Sass.

I think a Sass `variable` being consumed by a `placeholder` being consumed by an `@extend` being consumed by a `@mixin` can be simplified a little.

PR #736, opened 2022-04-29
introducing utility classes generated from design tokens

sturobson:
Now that we have the start of Design Tokens in the front-end toolkits there are places where we could use it to replace existing code that's either 'hand coded' or created using Sass. 

This PR _currently_ demonstrates how we can parse the Design Tokens `.json` and generate utility classes from those tokens and include the values and naming convention. 

## Elements Roadmap Things

- [ ] document how this is created, how this can be implemented.
- [ ] announcement of *new* utility classes in #ask-elements slack channel.
- [ ] deprecate the Sass function that currently generates the spacing utility classes.


sturobson:
I've merged the scripts to make this and documentation into #835 so we can close this.

PR #737, opened 2022-04-29
adds curly brace to missing design tokens value

sturobson:
@morgaan reached out with a but that was producing 

`$tokens--background-container: color.white};`

looks like we missed a curly bracket in a `.json` file. 

This PR fixes that.

To make sure this doesn't happen in future - I'm investigating ways to 'test' the `.json` so it adheres to some sort of spec. 

🙏🖤

### DIFF
--- a/context/brand-context/HISTORY.md
+++ b/context/brand-context/HISTORY.md
@@ -1,5 +1,9 @@
 # History
 
+## 21.0.1 (2022-04-29)
+    * BUG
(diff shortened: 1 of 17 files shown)


alexkilgour (approved):

sturobson:
Greg Miller in a plaid jacket pointing at the camera and saying 'thank you sir'

PR #738, opened 2022-04-29
manually removes `_index.scss` file from the `_index.scss` file

sturobson:
It appears my little bit of JS to generate a manifest file for each brands set of tokens has regressed and is now including itself in the list of `@imports` 

This PR is a 'hotfix' (I guess) as I've manually removed the `@import "_index.scss" for each brands `00-tokens`. 

I've run a local Sass compile and it's not erroring now. 

🙏🖤

### DIFF
--- a/context/brand-context/HISTORY.md
+++ b/context/brand-context/HISTORY.md
@@ -1,5 +1,8 @@
 # History
 
+## 21.0.2 (2022-04-29)
+    * BUG
+      * manually removes `_index.scss` file from the `_index.scss` file for each brands generated tokens.
 ## 21.0.1 (2022-04-29)
     * BUG
       * fixes issue with compiled Design Tokens having an missing opening curly brace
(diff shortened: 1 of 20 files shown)


alexkilgour (approved):
Approve fix. Is this something that needs a more permanent fix in the JS?

morgaan:
> … and I've just worked out why it was doing it.
> 
> The task to generate the aliases (background colours, text colours, etc) runs the literal task first.
> 
> Because some of the tokens generated are literal (spacing, breakpoints, sizing) that build script had the code to generate an index file.
> 
> So when the alias build task creates and index file it's already the file already exists and it overwrites with the additional `@import: _index.scss`.
> 
> So … any minute now there'll be a third PR (the second to resolve the `index.scss` file trying to eat itself.
> 
> And then, I'll have a lie down.

Not sure I've got it all, I'll have to dive into the ins and out of Design Tokens, but I'm glad you were able to spot the issue and about that you are about to ship a permanent fix. Respect!

Oh yes, well deserved lie down 🤗

PR #739, opened 2022-04-29
stops the `_index.scss` file from trying to import itself

sturobson:
There was an issue @morgaan spotted where each brands Design Tokens `_index.scss` manifest file was trying to import itself.

I quickly fixed this in #738 thinking it was a regression in the JavaScript that generated the file. 

It was actually because that JavaScript ran twice. 

The current tokens that generate the Sass variables are both 'literal' and 'alias' tokens. 

Both build scripts in the `design-tokens` folder ran JavaScript to generate an `_index.scss` from the files that are created in the folder.

These tasks are combined in the `scripts` in the package.json - which meant the `_index.scss` file generation was ran twice. The second time taking into account the existing `_index.scss` file. 

So - I've removed the JavaScript that generates an `index.scss` file for literal tokens and that fixes that.


sturobson:
In short: `literal` tokens that inform the `alias` tokens (and eventually they would inform `component` tokens).

So `literal` tokens will not generate Sass by on their own, but only as part of generating all the other tokens into Sass variables. 

Because of this removing the `index.scss` generation JavaScript from `build-literal.js` makes more sense than re-writing part of `build-alias.js` to not include the `_index.scss` file when creating itself.

morgaan (approved):
LGTM, thank you for fixing that and giving more context to the PR description ;)

PR #741, opened 2022-05-04
using default base size of 1.125rem

Heydon:
The furniture font size was incorrectly `1rem`. Since we hack the `body` size for an old Chrome bug it should be `1.125rem`, hence this PR!


sturobson (approved):
Mighty Morphing Power Rangers!

Issue #485, opened 2021-03-05
global-forms: master issue


alexkilgour:
This issue aims to link all the other `global-forms` issues together in order to facilitate discussion around how we should handle this going forward

## Issues

- #484 
- #168 
- #48 
- #49 
- #50 
- #51 

## Considerations

Some considerations for how we handle forms going forward

- Standardised UX across different brands?
- How do we handle error messaging (we already have `global-status-message`)?
- Handle form elements in a single component (`global-forms`) or individual ([like GDS](https://design-system.service.gov.uk/components/checkboxes/))?
- Do we include javascript validation?

sturobson:
Now that a reworked version of `global-forms` is up and running as a release candidate version:

1. are the issues linked above still relevant for todays use cases?
2. are they fixed with the current release candidate?
3. can they be worked on as part of the release candidate workflow for '2022' `global-forms`?

cc @alexkilgour @Heydon

sturobson:
There has been no update on this issue for 30 days. It is now marked as "stale". If you think this issue is still relevant and needs attention then reply with an update. If there is no update in 14 days, this issue will be closed.

PR #743, opened 2022-05-05
Global Forms 

Heydon:
Major reworking of `global-forms` following the formation of The Commitee For HTML Forms™️ (@alexkilgour, @Heydon, @sturobson, @axemonkey, @sky-jack). Design is informed by @foxintherain's work: https://www.sketch.com/s/c77d316e-76e5-403e-95b8-93021fc6c95e/p/0EB75A8C-3CC3-40C8-A48E-AE685CB267BC. Note that the button styles have _not_ been updated to reflect the Sketch file since this is a separate concern with a broader scope.

This is an RC (release candidate) release.

From the **HISTORY.md** file:

```
## 5.0.0-rc.1 (2022-06-27)
    * BREAKING:
        * Form fields now available as a suite of templates/partials
        * Major changes to styling and styling API
        * Design tokens support
```

Includes comprehensive documentation.


toolkits/global/packages/global-forms/view/inputCommonAttributes.hbs
```diff
@@ -0,0 +1,18 @@
+{{#if id}} id="{{id}}"{{/if}}
+{{#if name}} name="{{name}}"{{/if}}
+{{#if value}} value="{{value}}"{{/if}}
+{{#if error}} invalid aria-invalid="true"{{/if}}
+{{#if error}} aria-describedby="error-{{id}}"{{/if}}
```
aliuk2012, line 5: Should these be combined into a single `if` statement?

Heydon: Maybe? I but them on separate line for readability?

sturobson: I'm a fan of clarity over brevity to help folks understand what the code is doing. 

I think keeping them as two would help :)

context/brand-context/default/scss/00-tokens/_background.variables.scss
```diff
@@ -1,6 +1,6 @@
 
 // Do not edit directly
-// Generated on Tue, 28 Jun 2022 10:44:21 GMT
+// Generated on Wed, 29 Jun 2022 13:55:09 GMT
```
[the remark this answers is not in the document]
Heydon: @sturobson Would you mind adding that change since it is tokens territory?

toolkits/global/packages/global-forms/README.md
```diff
@@ -13,3 +15,265 @@ The `global-forms` component currently uses the `DEFAULT` brand only.
...
+
+Each field can have an `error` property. The presence of the property indicates the field is in an error state and the property value (a string) defines the error message.
+
+```
+{
+    "type": "text",
+    "label": "Your name",
+    "id": "your-name",
+    "name": "your-name",
+    "error": "That is not a name, it’s a number"
```
sturobson, line 121: more of a 'stub'. Does SN Digital have defined microcopy for things like form errors? I wonder if a) it's worth putting anywhere here (probably not) or b) making sure it's part of the Elements Design System site for folks to reference.

sturobson:
> Note that the button styles have not been updated to reflect the Sketch file since this is a separate concern with a broader scope.

Perhaps this is the change to rethink buttons, a component, and whether they should get upgraded to a `c-` component rather than a (multi-rule) `u-` utility class

Heydon:
> Perhaps this is the change to rethink buttons, a component, and whether they should get upgraded to a c- component rather than a (multi-rule) u- utility class

I agree we should head in that direction. Probably keep the utility classes and provide a component on top, letting one list an array of modifiers as I have done for the `buttons` field here. Separate PR, though, probably.

aliuk2012:
Looking at the demo,

image
Could we not override the error state if the user selects the checkbox? it looks really odd still having a thin slither of red border.

aliuk2012:
Could we add an example of what radio buttons with errors look like?

aliuk2012:
At 100% the inner circle of radio buttons look like filled in checkboxes with  rounded corners.
image

But I can see when zoomed in that it should be a circle 

image

context/brand-context/springernature/scss/00-tokens/_typography.variables.scss
```diff
@@ -1,6 +1,6 @@
 
 // Do not edit directly
-// Generated on Tue, 28 Jun 2022 10:44:21 GMT
+// Generated on Wed, 29 Jun 2022 13:55:09 GMT
 
 $tokens--typography-heading-font-family: serif;
```
aliuk2012, line 5: Should this not be using merriweather?

aliuk2012:
Should the demo of the components look exactly like the sketch file version? it seems like a few styles are missing from it. For example hint text is not grey and a few other things. https://jsfiddle.net/mnsL5pk7/

Heydon:
> Could we add an example of what radio buttons with errors look like?

Good question. I don't imagine a case where you could choose a wrong choice with radio buttons, but it's worth supporting it anyway.

aliuk2012:
> > At 100% the inner circle of radio buttons look like filled in checkboxes with rounded corners.
> 
> Looks like you ran out of pixels. This isn't a CSS issue but a rendering engine one. What browser is it, out of interest?

I'm using chrome Version 103.0.5060.53 (Official Build) (arm64). I remember seeing this issue before in the past but cant remember what ended up fixing it.

Heydon:
> Should the demo of the components look exactly like the sketch file version? it seems like a few styles are missing from it. For example hint text is not grey and a few other things. https://jsfiddle.net/mnsL5pk7/

Exactly? No, we're not pursuing "pixel perfection". But if you could list the disparities you have detected, that would be helpful. Well spotted on the gray hint text, I've just commited that change 👍

toolkits/global/packages/global-forms/view/fields/radios.hbs
```diff
@@ -0,0 +1,18 @@
+{{#if inputs}}
+	<fieldset class="c-forms__fieldset c-forms__field" {{#unless legend}} role="presentation"{{/unless}} {{> inputCommonAttributes}}>
```
aliuk2012, line 2: I don't think we can reuse `inputCommonAttributes` here because the fieldset then ends up with duplicate id to the the form field. I also spotted a similar issue on a `div`. If you look at the demo.html there are a few. 
eg in devtools console run `document.querySelectorAll('#textarea-example')` you will see two elements with the same id, a `div` and `textarea`. The textarea is a child element of the div

context/brand-context/default/scss/00-tokens/_typography.variables.scss
```diff
@@ -0,0 +1,22 @@
+
+// Do not edit directly
+// Generated on Wed, 29 Jun 2022 21:08:16 GMT
+
+$tokens--typography-heading-font-family: serif;
+$tokens--typography-heading-letter-spacing: 0.2rem;
+$tokens--typography-heading-line-height: 1.2;
+$tokens--typography-heading-level-1-default-font-size: 2.5rem;
+$tokens--typography-heading-level-1-default-font-weight: 400;
+$tokens--typography-heading-level-1-fluid-font-size: clamp(1.875rem, 4vw, 2.5rem);
```
morgaan, line 10: What is the fallback for browsers not supporting clamp()? Browser default styles or do we have some sensible defaults?

context/design-tokens/build-component.js
```diff
@@ -18,40 +18,61 @@ function getStyleDictionaryConfig(brand, components) {
...
 	}
 }
-// TODO: see if I can combine this with an existing Sass file
 
 console.log('Build started...');
 
-['global'].map(function (brand) {
-	let dir = `${__dirname}/component/${brand}`
+// the following array needs to be manually updated when more brand speicifc components are added to brands not on this list. Otherwise this will not build. This is why it is currently not part of
+['springer'].map(function (brand) {
```
morgaan, line 52: Does this means that so far `global-forms`, if not all design-tokens aware components, only supports the springer brand?

Sorry if this question sounds weird, I've no experience so far with design-token and have too less of an overview of this big rehaul work

context/brand-context/default/scss/00-tokens/_typography.variables.scss
sturobson: this isn't really part of this PR, as the design tokens were merged some time ago … 

however, the expectation is that these values are implemented (kind of) like [they are for Nature](https://github.com/springernature/frontend-toolkits/blob/master/context/brand-context/nature/scss/30-mixins/typography.scss#L15) (although, I'd suggest not as a variable in a placeholder that's used in a mixin, that's a bit Sass overkill). 

I guess the fallback would be [similar [to that in Springer Nature[(https://github.com/springernature/frontend-toolkits/blob/master/context/brand-context/springernature/scss/30-mixins/typography.scss#L3) where with no fallback provided … an non-supporting browser would (presumably) default to the user agent. 

This is why there is the `default` value and the `fluid` value - so there's a fallback. 

I'll try and work out how to document this to be more explicit (getting Tokens documentation genearated as part of the documentation on the Elements site would help)

context/design-tokens/build-component.js
sturobson: this might be my lack of JS skills -- but if I include all brands but a `.json` file for that that brand doesn't exist - the process ends. 

So, I'm aware of this, and in lieu of better JavaScript skills I'll add brands as and when brand specific component tokens are available :)

toolkits/global/packages/global-forms/README.md
```diff
@@ -13,3 +15,265 @@ The `global-forms` component currently uses the `DEFAULT` brand only.
...
+
+```
+{
+    "type": "radios",
+    "label": "Animal",
+    "inputs": [
+        {
+            "label": "Horse",
+            "value": "Horse",
+            "name": "animal"
```
morgaan, line 172: Would this make sense to hoist the name property from the inputs items to the field properties. Less repetitive and maybe less prone to mistakes?

Heydon: @morgaan I attempted that but handlebars wasn't having it for some reason. I might have another try.

sturobson: "handlebars – there's probably a helper you need to write for that"

toolkits/global/packages/global-forms/package.json
```diff
@@ -1,12 +1,12 @@
...
 	"license": "MIT",
 	"description": "form component",
 	"keywords": [
 	  "css",
 	  "component"
 	],
   "author": "Jack Watkins",
-  "brandContext": "^18.0.0"
-  }
+  "brandContext": "^24.0.0"
```
morgaan, line 11: Should not this be 25 rather?

toolkits/global/packages/global-forms/README.md
morgaan: @Heydon Have you tried `../name`? So far I remember you can use this `..` operator to climb up the context.

toolkits/global/packages/global-forms/scss/50-components/_forms.scss
```diff
@@ -1,78 +1,236 @@
...
+	border: 0;
+}
+
+.c-forms__input,
+.c-forms__error-summary,
+.c-forms__input-container {
+	max-width: $global-forms-input-width-maximum;
+}
+
+:is(.c-forms__field, .c-forms__fieldset) + :is(.c-forms__field, .c-forms__fieldset) {
```
morgaan, line 40: Wow just found out about this amazing pseudo class I may have been dreaming of for years.
Is this safe to use, support-wise?

toolkits/global/packages/global-forms/view/fields/radios.hbs
Heydon: Ah, good shout. I am still getting used to the way handlebars deals in inheritance (I prefer nunjucks). Plus, Firefox a11y inspector apparently doesn't identify duplicate `id`s. We don't actually need `id`s on the fields `<div>`s, I just put them there for scripting convenience. This should be fixed now.

toolkits/global/packages/global-forms/package.json
Heydon: @sturobson  Just done so! https://github.com/springernature/frontend-toolkits/pull/743/commits/227a9228fda6bfcce9109db7464bd4fa268204ba

toolkits/global/packages/global-forms/scss/50-components/_forms.scss
Heydon: @morgaan Come to think of it, I'm not sure our position on IE about this currently. Have rolled back for the time being. https://github.com/springernature/frontend-toolkits/pull/743/commits/28a443419b5c5b1d538ef59863c91e9f7c307d84

toolkits/global/packages/global-forms/README.md
Heydon: @morgaan It is a bit tricky with partials but I got it working with `{{> radio name=../name }}`. Updated the **README** to match. https://github.com/springernature/frontend-toolkits/pull/743/commits/4080e47d08be4220d95554454978c8dde53522ec

toolkits/global/packages/global-forms/scss/00-tokens/_default.tokens.scss
```diff
@@ -0,0 +1,46 @@
+// Generated on 30/06/2022, 08:09:54
```
benjclark, line 1: Am I right in thinking lots of these variables are not being used in the component styling? From what I can tell these are defined but not used:

$global-forms-checkbox-checked-icon-fill
$global-forms-checkbox-checked-icon-path
$global-forms-checkbox-checked-icon-stroke
$global-forms-checkbox-size
$global-forms-heading-font-family
$global-forms-heading-font-size
$global-forms-heading-font-weight
$global-forms-indent-border-radius
$global-forms-indent-border-width
$global-forms-input-border-color-error
$global-forms-input-font-color
$global-forms-label-font-size
$global-forms-legend-font-family
$global-forms-legend-font-size
$global-forms-legend-font-weight
$global-forms-message-color-resting
$global-forms-message-font-size
$global-forms-message-font-weight
$global-forms-message-icon-color-error
$global-forms-message-icon-color-resting
$global-forms-radio-carat-color
$global-forms-radio-carat-size
$global-forms-radio-carat-spacing
$global-forms-select-icon-stroke

Perhaps I'm missing something? 😄

toolkits/global/packages/global-forms/README.md
```diff
@@ -13,3 +15,265 @@ The `global-forms` component currently uses the `DEFAULT` brand only.
...
+        },
+        {
+            "id": "checkbox-terms-error",
+            "error": "You must agree to the terms!"
+        }
+    ]
+}
+```
+
+### Multiple choices
```
morgaan, line 144: Hold on, radios and select (unless using `multiple` attribute) are not multiple choices. Do we support the `select[multiple]` so far?
Shall we host these in a different section?

toolkits/global/packages/global-forms/view/fields/email.hbs
```diff
@@ -0,0 +1,7 @@
+{{> label}}
+{{#if error}}
+	{{> error}}
+{{/if}}
+<input type="email" class="c-forms__input c-forms__input--email"
```
aliuk2012, line 5: why do we need separate templates for different types of inputs? couldn't we just pass in the type that we would like to use and by default inputs are set to text

toolkits/global/packages/global-forms/view/fields/radios.hbs
aliuk2012: Maybe not for this PR directly but in future it would be good to use Pally or aXe on the component in its various states to see if theres any accessibility issues from a component level. We did a similar thing at GDS which worked well. Happy to talk you through it.

toolkits/global/packages/global-forms/scss/00-tokens/_default.tokens.scss
Heydon: @benjclark No, you're quite right. @StuRobson and I need to audit what should be used and what's redundant.

toolkits/global/packages/global-forms/view/fields/email.hbs
morgaan: Maybe because of `c-forms__input--email`

morgaan: unless we style it using `[type="email"]`

Heydon: @aliuk2012 Not really, because component selection is already dynamic and some inputs are actually selects, textarea, etc.

toolkits/global/packages/global-forms/README.md
Heydon: @morgaan Terminology clash. I mean each of these elements represent multiple choices (e.g. options for selects).

toolkits/global/packages/global-forms/scss/00-tokens/_default.tokens.scss
sturobson: it's also a case of me being explicit in trying to define everything … and @Heydon (perhaps rightly (😛), considering the overall architecture of the CSS) relying on the cascade

benjclark:
I can see that the majority of the css property values in `toolkits/global/packages/global-forms/scss/50-components/_forms.scss` reference `$global-forms-` namespaced Sass variables which are ultimately generated from a tokens json file for that component: `context/design-tokens/components/global/global-forms/default.json`.

I'm just wondering how essential it is to have tokens json files defined for every component?

(forgive me, I'm no doubt being incredibly naive) 😄 

If the alias and literal tokens generated Sass variables that lived in brand context then components could simply reference those in a 10-settings scss files and it would remove the need for us to write token json files for every component. Which seems like a decent effort saving.

Is it because we need a component's css values to be in defined in token json format if we want them to be consumed by designer applications like Sketch and Figma? and if so, does that mean we need all of the component's css property values to be defined in tokens? Noted that some of the values you have defined as tokens, some you have not - where do you draw the line?

Thanks for this PR. Great stuff and so interesting to explore 👏

toolkits/global/packages/global-forms/README.md
morgaan: Oh I see, well I'm sure @amyhupe could help there. As a non-native speaker I won't be of any help here I'm afraid. Lot of luck in finding better copy.

toolkits/global/packages/global-forms/view/fields/email.hbs
aliuk2012: if `<input...` was made generic then then that view template could be reused for radio/checkboxes/file and all the other various type of inputs.

toolkits/global/packages/global-forms/README.md
morgaan: 💯 just as well as in general when under 5-6 options (Burn you select tags talk resonate in my head)

morgaan: This kind of guideline should make it to the doc actually in "When to use..." or "When not to use..."

sturobson:
> @benjclark Is it because we need a component's css values to be in defined in token json format if we want them to be consumed by designer applications like Sketch and Figma?

Yes, hopefully. I've managed to get this working with Figma (both 'pulling' and 'pushing' the json), as James was doing some research around tooling with Design … but as Design (are still currently) use Sketch there maybe some time to wait, as Sketch is lagging behind a little here.

It also allows integration of the component token information to be generated for a components documentation page in the Elements Design System site. 

> If the alias and literal tokens generated Sass variables that lived in brand context then components could simply reference those 

Abstracting to these three levels, as well as the future benefits above, I feel, should also close the gap between Design and Development a little more _and_ improve consistency across brands and products. 

> Does that mean we need all of the component's css property values to be defined in tokens?

As I've [mentioned here](https://github.com/springernature/frontend-toolkits/pull/743#discussion_r910917121) I've a tendency towards being completely explicit and forgoing the cascade … but as the existing architecture makes great use of this I think we can pair down 'component tokens' to accommodate it. Me and @Heydon have added 'paring down the tokens' to the final todo list for this PR. 

> where do you draw the line?

Still a work in progress 😆

toolkits/global/packages/global-forms/scss/00-tokens/_default.tokens.scss
```diff
@@ -0,0 +1,34 @@
+// Generated on 30/06/2022, 16:20:58
+// Source: design-tokens/componenets/global/global-forms/default.json
```
alexkilgour, line 2: Spelling mistake: `componenets` should be `components`
across this and springer

sturobson: whoops! 

I had a bit of a re-think on this, as an aside. 

The current `npm scripts` for generating the Sass variables are a pretty blunt tool in generating `literal` and `alias` tokens (pretty much) every time. 

I'm going to rework it so that they're either _very_ separate tasks or the script checks for changes to disc for the`.json` files and only compiles things that have changed. 

Moving forward, it should be the case that `literal tokens` very rarely get changed. `alias tokens` do get changed but not often, and `component` tokens` only get changed/added to when there are changes from the `alias` tokens or we are introducing `component` tokens to the component.

If that makes sense.

sturobson:
with the linting failing on shortened hex color codes (or them not being shortened) hadn’t failed when they were introduced in the brand-context - This issue should soon be irrelevant when we switch back on the output to reference the alias token Sass variables.

toolkits/global/packages/global-forms/README.md
```diff
@@ -13,3 +17,265 @@ The `global-forms` component currently uses the `DEFAULT` brand only.
...
+        "horse",
+        "emu",
+        "platypus",
+        "cockroach",
+        "whale"
+    ]
+}
+```
+
+This builds a standard `<datalist>` element—with `<option>`s—and associates it with the input. If a `<datalist>` already exists in the markup, provide just the `id` and forego the `options` property. If you wish to implement a bespoke type ahead solution, using JavaScript, omit the `datalist` property from the field altogether.
```
amyhupe, line 281:
```suggestion
This builds a standard `datalist` field, with `option`s, and associates it with the input. If a `datalist` already exists in the markup, provide just the `id` and forego the `option` property. If you want to use a custom-built autocomplete solution using JavaScript, omit the `datalist` property from the field altogether.
```

toolkits/global/packages/global-forms/demo/context.json
```diff
@@ -0,0 +1,234 @@
...
+	"form1": {
+		"errorIconURL": "../img/error.svg#i-error",
+		"selectIconURL": "../img/chevron-more.svg#i-chevron-more",
+		"errorSummary": {
+			"id": "summary",
+			"title": "Here’s what’s wrong",
+			"errors": [
+				{
+					"id": "your-name-error",
+					"error": "That is not a name, it’s a number"
```
amyhupe, line 11:
```suggestion
					"error": "Enter your name"
```

toolkits/global/packages/global-forms/demo/context.json
```diff
@@ -0,0 +1,234 @@
...
+				},
+				{
+					"id": "checkbox-terms-error",
+					"error": "You must agree to the terms before continuing"
```
amyhupe, line 15: To make consistent with my previous suggestion

```suggestion
					"error": "Agree to the terms to continue"
```

toolkits/global/packages/global-forms/demo/context.json
```diff
@@ -0,0 +1,234 @@
...
+						"template": "checkbox",
+						"label": "Agree to our terms (with hint)",
+						"hint": "You should have read these. Read them, then agree to them.",
+						"id": "checkbox-terms-hint",
+						"name": "checkbox-terms-hint"
+					},
+					{
+						"template": "checkbox",
+						"label": "Agree to our terms (with error)",
+						"error": "You must agree to the terms before continuing.",
```
amyhupe, line 177: Updating to keep consistent with my previous suggestions:

```suggestion
						"error": "Agree to the terms to continue.",
```

toolkits/global/packages/global-forms/demo/context.json
```diff
@@ -0,0 +1,234 @@
...
+					{
+						"template": "checkbox",
+						"label": "Agree to our terms (basic)",
+						"id": "checkbox-terms",
+						"name": "checkbox-terms"
+					},
+					{
+						"template": "checkbox",
+						"label": "Agree to our terms (with hint)",
+						"hint": "You should have read these. Read them, then agree to them.",
```
amyhupe, line 170: Changing to something I'd be happy to see in a real service

```suggestion
						"hint": "You must agree with the terms to continue your application",
```

toolkits/global/packages/global-forms/demo/context.json
```diff
@@ -0,0 +1,234 @@
...
+						"id": "checkbox-terms-hint",
+						"name": "checkbox-terms-hint"
+					},
+					{
+						"template": "checkbox",
+						"label": "Agree to our terms (with error)",
+						"error": "You must agree to the terms before continuing.",
+						"hint": "You should have read these. Read them, then agree to them.",
```
amyhupe, line 178: As before

```suggestion
						"hint": "You must agree with the terms to continue your application.",
```

toolkits/global/packages/global-forms/README.md
```diff
@@ -13,3 +17,265 @@ The `global-forms` component currently uses the `DEFAULT` brand only.
 // Include this with your other components
 @import '@springernature/global-forms/scss/50-components/forms';
 ```
+
+Then you will need to register the Handlebars partials in the `/view` folder.
```
amyhupe, line 21: For consistency: 

```suggestion
Then you will need to register the handlebars partials in the `/view` folder. 
```

amyhupe (commented):
OK, here's my review. Mostly small things, aimed at:

- increasing consistency of terminology, as much as possible
- referencing user impact, where appropriate
- simplifying language

I've also suggested copy changes to some of the example copy we're using. My general take on this is that example copy should be obviously an example (e.g. Monkey, Platypus etc.) or realistic, and something we'd be happy to be used in a real service. 

I hope these comments make sense - let me know if anything needs clarifying. And good work - this is an absolute beast!

toolkits/global/packages/global-forms/README.md
```diff
@@ -13,3 +17,265 @@ The `global-forms` component currently uses the `DEFAULT` brand only.
 // Include this with your other components
 @import '@springernature/global-forms/scss/50-components/forms';
 ```
+
+Then you will need to register the Handlebars partials in the `/view` folder. 
+
+There are two options for rendering form fields: 
+
+1. Compile each field independently, using its respective template from the `view/fields` folder.
+2. Organize your fields into fieldsets using the data structure exemplified in `demo/context.json`. In this case, you will have to iterate over a `fieldsets` property something like this:
```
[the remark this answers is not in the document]
Heydon: Putting"over a fieldsets array" to make it more explicit what the iteration refers to.

toolkits/global/packages/global-forms/README.md
```diff
@@ -13,3 +17,265 @@ The `global-forms` component currently uses the `DEFAULT` brand only.
...
+		{{#each fieldsets}}
+			{{> fieldset }}
+		{{/each}}
+    {{/with}}
+</form>
+```
+
+### Fieldsets
+
+Fieldsets are used to group fields itemized under their `fields` property. If you do not want to include a (visible; screen reader identifiable) fieldset element or `<legend>`, simply omit the `legend` property. The following example represents a simple form body with a single, unlabeled fieldset containing two text inputs:
```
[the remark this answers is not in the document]
Heydon: Yeah, it doesn't need to be the element code here.

toolkits/global/packages/global-forms/README.md
```diff
@@ -13,3 +17,265 @@ The `global-forms` component currently uses the `DEFAULT` brand only.
...
+            "legend": "<h2>My level 2 legend</h2>",
+            "fields": [
+                ...
+            ]
+        }
+    ]
+}
+```
+
+(**Note:** it is valid and conforming to use heading elements inside `<legend>`s. For visual and accessible hierarchy it is strongly recommended.)
```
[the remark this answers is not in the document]
Heydon: I'll make the preceeding paragraph stronger about using headings and lose this.

toolkits/global/packages/global-forms/README.md
```diff
@@ -13,3 +17,265 @@ The `global-forms` component currently uses the `DEFAULT` brand only.
...
+{
+    "template": "text",
+    "label": "Your name",
+    "id": "your-name",
+    "name": "your-name",
+    "error": "That is not a name, it’s a number"
+}
+```
+
+Errors can be summarized using a top level `errorSummary` property (adjacent to the `fieldsets` property). Each error in the errors array must point to the `id` of the respective input and repeat its `error` message:
```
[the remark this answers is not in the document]
Heydon: I'm saying `input` because I would expect most specific errors on (text) input errors, but that is indeed misleading.

toolkits/global/packages/global-forms/README.md
```diff
@@ -13,3 +17,265 @@ The `global-forms` component currently uses the `DEFAULT` brand only.
...
+
+```json
+"errorSummary": {
+    "id": "summary",
+    "title": "Here’s what’s wrong",
```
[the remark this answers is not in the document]
Heydon: Putting "There are problems" for multiple errors

toolkits/global/packages/global-forms/README.md
```diff
@@ -13,3 +17,265 @@ The `global-forms` component currently uses the `DEFAULT` brand only.
...
+            "id": "checkbox-terms-error",
+            "error": "You must agree to the terms!"
+        }
+    ]
+}
+```
+
+### Making choices
+
+Select elements supply their choices via an `options` property, which must be an array. The `selected` property is a Boolean:
```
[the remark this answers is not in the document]
Heydon: This would be the only place the element is in code backticks but without `<` and `>` I think.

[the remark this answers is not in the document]
Heydon: Ah, see my previous comment.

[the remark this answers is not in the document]
Heydon: Hmm. I don't like "nest the user's choices"—nesting implies multiple levels of hierarchy. Hopefully, this is okay:

_Select fields (using the `<select>` element) define their choices via an `options` property, which must be an array._

toolkits/global/packages/global-forms/README.md
sturobson: as a long time owner of Sassnotsass.com I had a quick look at the docs for handlebars … and although the logo is in lowercase they seem to use Title Case when it's mentioned throughout the docs. 

(I'm happy either way 😄)

toolkits/global/packages/global-forms/README.md
```diff
@@ -13,3 +17,265 @@ The `global-forms` component currently uses the `DEFAULT` brand only.
...
+    },
+    {
+        "label": "Platypus",
+        "value": "Platypus",
+        "selected": true
+    }
+]
+```
+
+Radios provide _their_ choices via an `inputs` array:
```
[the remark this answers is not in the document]
Heydon: _Radios define choices via an `inputs` array_

toolkits/global/packages/global-forms/README.md
```diff
@@ -13,3 +17,265 @@ The `global-forms` component currently uses the `DEFAULT` brand only.
...
+            "id": "checkbox-horses",
+            "name": "checkbox-horses"
+        }
+    ]
+}
+```
+
+### Buttons
+
+A `template` of `buttons` defines a set of button controls, displayed inline (using Flexbox and `gap` for tidy wrapping).
```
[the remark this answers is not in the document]
Heydon: It's at a higher level than the utility, using the utility but in a way that fits with the fieldsets/fields paradigm. I'll make a note like so:

_The `modifiers` property is an array. Each value should match one of these modifiers `form the brand context’s button utility styles`._

toolkits/global/packages/global-forms/README.md
```diff
@@ -13,3 +17,265 @@ The `global-forms` component currently uses the `DEFAULT` brand only.
...
+            "name": "checkbox-horses"
+        }
+    ]
+}
+```
+
+### Buttons
+
+A `template` of `buttons` defines a set of button controls, displayed inline (using Flexbox and `gap` for tidy wrapping). 
+The `type` property for each individual button corresponds to the standard `type` property/attribute. For example, here is how you would include a submit button:
```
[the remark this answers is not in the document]
Heydon: I've just removed "standard", I don't think it's needed.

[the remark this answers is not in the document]
Heydon: Confusion through overcommunication.

toolkits/global/packages/global-forms/README.md
```diff
@@ -13,3 +17,265 @@ The `global-forms` component currently uses the `DEFAULT` brand only.
...
+                "type": "submit",
+                "label": "Submit",
+                "modifiers": ["primary"]
+            }
+        ]
+    }
+]
+```
+
+The `modifiers` property is an array. Each value should match one of these modifiers form the brand context:
```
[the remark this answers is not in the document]
Heydon: See https://github.com/springernature/frontend-toolkits/pull/743#discussion_r913129487

sturobson:
@alexkilgour I think this PR … would be a good chance to clean up the issues list … as I spotted this last night -- https://github.com/springernature/frontend-toolkits/issues/485 -- I'm assuming some of these would be nullified with this PR?

toolkits/global/packages/global-forms/view/label.hbs
```diff
@@ -0,0 +1,6 @@
+<label for="{{id}}" class="c-forms__label">
+	{{label}}{{#if optional}} (optional){{/if}}
```
aliuk2012, line 2: 🤔  I don't think "optional" is really helpful for the following reasons

1. By default all text fields are "optional" i.e they dont have "required" set to true.
2. This will not be useful if the service has to support multiple languages because the copy is not translated.

When I worked in GOV we tried to make all fields required by default (no point asking a user for data if its not useful for all) and only flag up "optional"  fields in the label like what is done here but in our case the text fields are not required by default so seems to send mix messages as to when to use this option or not.

aliuk2012: Instead just let the devs control the copy and if they want to use "(Optional)" they could just include it in their copy that they need for the label.

sturobson: The determination of what is optional or not, or what copy should be used would come from agreement across dsign/ux/content that would ultimately end up in the Design System docs site?

amyhupe: >Instead just let the devs control the copy and if they want to use "(Optional)" they could just include it in their copy that they need for the label.

From a content design point of view, I'd really like to introduce some standardisation if possible. I get that we're not currently offering localisation of this copy, but we're not currently doing that for the content elements of any components so this isn't a new problem. My concern is if we don't include something, people are going to fall back to using asterisks.

sturobson: So, @Heydon has followed the Sketch designs (see screenshot). I'd have to check with James (on holiday until Monday (11/07/22) but I think all of the form designs have gone through rounds of review in SN digital and this is currently the latest agreed across discipline.

Screenshot 2022-07-05 at 11 45 12

sturobson: regarding the issue with translation. It could be made a handlebars variable in the json, rather than hardcoded as it is now‽ So it could get switched out as needed? (pinging @Heydon) 

[Microcopy](https://www.microcopy.me) like (optional) here could also be tokenised -- which could further help documentation of it all

toolkits/global/packages/global-forms/README.md
```diff
@@ -13,3 +17,265 @@ The `global-forms` component currently uses the `DEFAULT` brand only.
...
+                ...
+            ]
+        }
+    ]
+}
+```
+
+### Fields
+
+Fields take the `template` property to determine which of the `view/fields` template files they represent. This property along with `label`, `id`, and The `template` property sets the type of field - for example, `"template": "text"` renders a text input field. Aim to make the `template`, `label`, `id`, and `name` properties mandatory parts of your data schema.
```
amyhupe, line 85: Something went awry with the resolving of these edits. Fixed:

```suggestion
The `template` property sets the type of field - for example, `"template": "text"` renders a text input field. Aim to make the `template`, `label`, `id`, and `name` properties mandatory parts of your data schema.
```

toolkits/global/packages/global-forms/README.md
```diff
@@ -13,3 +17,265 @@ The `global-forms` component currently uses the `DEFAULT` brand only.
...
+
+This component supports a wide range of standard form field attributes. So, for example, to include a `readonly` attribute on your text input, you  can include a property of the same name on the data:
```
amyhupe, line 87:
```suggestion
This component supports a wide range of standard form field attributes. For example, to include a `readonly` attribute on your text input, you can include a property of the same name on the data:
```

toolkits/global/packages/global-forms/README.md
```diff
@@ -13,3 +17,265 @@ The `global-forms` component currently uses the `DEFAULT` brand only.
...
+{
+    "template": "text",
+    "label": "Your name",
+    "id": "your-name",
+    "name": "your-name",
+    "error": "Enter your name"
+}
+```
+
+You can summarise errors using a top level `errorSummary` property (adjacent to the `fieldsets` property). Each error in the errors array must point to the `id` of the input it relates to and repeat its `error` message:
```
amyhupe, line 125:
```suggestion
You can summarise errors using a top level `errorSummary` property adjacent to the `fieldsets` property. Each error in the errors array must point to the `id` of the input it relates to and repeat its `error` message:
```

toolkits/global/packages/global-forms/README.md
```diff
@@ -13,3 +17,265 @@ The `global-forms` component currently uses the `DEFAULT` brand only.
...
+            "id": "checkbox-terms-error",
+            "error": "Agree to the terms to continue"
+        }
+    ]
+}
+```
+
+### Making choices
+
+`Select` fields (using the `<select>` element) define their choices via an `options` property, which must be an array. The `selected` property is a Boolean:
```
amyhupe, line 146: I think it's good to specify what "their choices" refer to. I know this is technical documentation, but the more we can do to contextualise it for designers now, the less we'll have to edit later when we layer in the design guidelines. 

```suggestion
`Select` fields (using the `<select>` element) define the user's answer choices with an `options` property, which must be an array. The `selected` property is a Boolean:
```

toolkits/global/packages/global-forms/README.md
```diff
@@ -13,3 +17,265 @@ The `global-forms` component currently uses the `DEFAULT` brand only.
...
+        }
+    ]
+}
+```
+
+The `name` property is placed at the top level and inherited by each input.
+
+Radios are always grouped together in a `fieldset`, so the group label (“Animal”, here) renders as a `legend` not a `label`.
+
+You might want to show users an additional field when they select a radio. For example, revealing a text input field for them to give more specific information about the option they’ve selected.
```
amyhupe, line 187:
```suggestion
You might want to show users an additional field when they select a particular radio. For example, revealing a text input field for them to give more specific information about the option they’ve selected. 
```

toolkits/global/packages/global-forms/README.md
```diff
@@ -13,3 +17,265 @@ The `global-forms` component currently uses the `DEFAULT` brand only.
...
+            "label": "What type of monkey?",
+            "id": "monkey-type",
+            "name": "monkey-type",
+            "hint": "Chimps are not monkeys"
+        }
+    ]
+}
+```
+
+Unlike radios, which are always used in sets of two or more, you can have a single checkbox field. To give users a set of checkbox choices, organise the checkboxes into a `fieldset`:
```
amyhupe, line 208: I think we can tie these 2 sentences together a bit better, like this:

```suggestion
Unlike radios, which are always used in sets of two or more, you can have a single checkbox field. However, if you need to give users a set of checkbox choices, organise the checkboxes into a `fieldset`:
```

toolkits/global/packages/global-forms/README.md
```diff
@@ -13,3 +17,265 @@ The `global-forms` component currently uses the `DEFAULT` brand only.
...
+            "name": "checkbox-horses"
+        }
+    ]
+}
+```
+
+### Buttons
+
+A `template` of `buttons` defines a set of button controls, displayed inline (using Flexbox and `gap` for tidy wrapping). 
+The `type` property for each individual button corresponds to the `type` property/attribute. For example, here is how you would include a submit button:
```
amyhupe, line 233: Can we make a choice here between property and attribute, and lose the slash? If it really could be either, can we say "or"?

amyhupe (commented):
Hello,

A few additional revisions on yesterday's revisions. These are things I think could be better clarified, on a second read, or where copy errors have been introduced in the process of suggesting & accepting changes.

toolkits/global/packages/global-forms/view/fields/email.hbs
Heydon: No it couldn't, because radios are together represent one field.

toolkits/global/packages/global-forms/README.md
```diff
@@ -13,3 +17,265 @@ The `global-forms` component currently uses the `DEFAULT` brand only.
 // Include this with your other components
 @import '@springernature/global-forms/scss/50-components/forms';
 ```
+
+Then you will need to register the handlebars partials in the `/view` folder.
```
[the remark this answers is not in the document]
Heydon: You can, in a later version.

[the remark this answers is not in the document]
sturobson: > Could we include a js/java example here on how best to register the partials?

This sounds like an implementation detail that would be better documented further up in the Elements docs site, or in the application docs itself, or even somewhere in the Engineering Enablement rather than repetition per component docs page.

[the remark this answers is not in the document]
aliuk2012: 👍

toolkits/global/packages/global-forms/view/label.hbs
Heydon: I was just following the UX. Can we discuss this after v1 is out, please? For the time being, just put `required: true` on all your fields if you want them all required.

sturobson (approved):
"It's been a long road. Getting from there to here. It's been a long time. But the time is finally near."

I think we're now in a good place to release this as an `rc1` version.

Buyer Beware! There may be dragons!

Great work @Heydon and all involved in recreating `global-forms` for 2022 and beyond, finding bugs, questioning and improving things so `rc1` is pretty solid out of the gate.

PR #747, opened 2022-05-13
experimental integration of Figma Design Tokens plugin code 

sturobson:


sturobson:
Sadly closing this as the potential move from Sketch to Figma in the Design and UX disciplines isn't moving forward over costs.

PR #751, opened 2022-05-23
uses negation to left align lists without classes

Heydon:
```
## 21.1.1 (2022-05-23)
    * Left aligns basic lists in the springer context
```

`0.9em` feels magic numbery 😬 but I tested across multiple fonts and font-sizes and it is the most reliable value.

```
/* Basic lists should be aligned to the left */
ul:not([class]),
ol:not([class]) {
	/* Allow for bullet itself */
	padding-left: 0.9em;
}
```

### DIFF
--- a/context/brand-context/HISTORY.md
+++ b/context/brand-context/HISTORY.md
@@ -1,4 +1,6 @@
 # History
+## 21.1.1 (2022-05-23)
+    * Left aligns basic lists in the springer context
 ## 21.1.0 (2022-05-10)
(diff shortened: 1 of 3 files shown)


sturobson (approved):
> 0.9em feels magic numbery

😆 

Looks good. This would be something that'd be nice to incorporate into the long-form content component (when I get to it). 

As it doesn't break functionality (really), and it's an old typographic thing from the dark ages of books I think we can merge this in without effecting other teams that use this `.scss` file.

LGTM!

PR #752, opened 2022-05-30
IE image fix

Heydon:
* The fallback for IE did not work quite as expected in real IE11
* This has been addressed by removing a `transform` and setting an explicit `left: 0` for the positioning

### Before

broken layout

### After

unbroken layout


sturobson (approved):
lovely stuff!

PR #761, opened 2022-06-29
Update nature hero

ricky-lau:
* Update brand-context (Node Sass to Dart Sass)
* Removes dependency on utility classnames

### DIFF
--- a/toolkits/nature/packages/nature-hero/view/hero.hbs
+++ b/toolkits/nature/packages/nature-hero/view/hero.hbs
@@ -1,4 +1,4 @@
-<div class="c-hero{{#if flushMdMax}} c-hero--flush-md-max{{/if}} u-mb-16 u-position-relative">
+<div class="c-hero{{#if flushMdMax}} c-hero--flush-md-max{{/if}}">
 	<div class="c-hero__image">
 		<picture>
 			<source type="{{picture.type}}" srcset="{{picture.srcset}}" sizes="{{picture.sizes}}">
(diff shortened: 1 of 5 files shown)


toolkits/nature/packages/nature-hero/view/hero.hbs
```diff
@@ -7,8 +7,8 @@
 	</div>
 	<div class="c-hero__copy">
 		<h2 class="c-hero__title">
-			<a class="c-hero__link u-link-faux-block" href="{{title.uri}}" data-track="click" data-track-action="{{title.dataTrackAction}}" data-track-label="link">{{title.text}}</a>
+			<a class="c-hero__link" href="{{title.uri}}" data-track="click" data-track-action="{{title.dataTrackAction}}" data-track-label="link">{{title.text}}</a>
 		</h2>
```
aliuk2012, line 11: I noticed in the styling that `c-hero__link`  is setup to be display:block. Would it not be better to wrap the <a> around the h2 because by default h2 are blocks and therefore the link wouldn't need to have is display attribute altered. Less CSS and just embracing html natural behaviour.

aliuk2012: @ricky-lau I noticed you pushed the changes to remove the display block on the link. Did you want the link to be clickable anywhere along the full width of the heading? If so then you will need to wrap the <h2> with <a> instead of having it inside the heading.

ricky-lau: The mixin `u-link-faux-block` is making the `c-hero` element clickable so I think the display block is redundant on the link.

aliuk2012: Oh I see that makes sense. 

🤔  If the whole hero needs to be clickable why do we need `u-link-faux-block` again could just wrap the `<a>` around `c-hero` div?

Sorry for all the questions just trying to understand our design system a bit more and the different components.

Everything else looks great in this PR

PR #764, opened 2022-07-11
Global forms rc2

Heydon:
```
## 5.0.0-rc.2 (2022-07-11)
    * BREAKING:
        * Arbitrary attributes now restricted to data attributes
        * All templates now namespaced so partials can be more safely registered
    * Documenting the need for importing button utilities
```


toolkits/global/packages/global-forms/README.md
```diff
@@ -48,13 +51,13 @@ Fieldsets are used to group fields itemised under their `fields` property. If yo
...
                 {
-                    "template": "text",
+                    "template": "globalFormText",
                     "label": "Your name",
                     "id": "your-name",
                     "name": "your-name"
                 },
                 {
-                    "template": "text",
+                    "template": "globalFormText",
```
sturobson, line 60: does this need to be `globalFormEmail` or does it not matter?

sturobson: similarly with all other instances? I guess it would be know what it is doing depending on the label, id, name in the `json` to negate each field needing it's own `globalFormThing`

toolkits/global/packages/global-forms/HISTORY.md
```diff
@@ -1,5 +1,11 @@
 # History
 
+## 5.0.0-rc.2 (2022-07-11)
+    * BREAKING:
+        * Arbitrary attributes now restricted to data attributes
+        * All templates now namespaced so partials can be more safely registered
+    * Documenting the need for importing button utilities
+
```
sturobson, line 8: updates to version `25.0.0` of brand-context?

toolkits/global/packages/global-forms/demo/context.json
```diff
@@ -84,7 +83,7 @@
 						]
 					},
 					{
-						"template": "select",
+						"template": "globalFormSelect",
 						"label": "Animal (with hint)",
```
sturobson, line 87: I think, at some point, we may want to use 'real world' SN content for this. I recall @amyhupe discussing it in a wider Elements ideal.

sturobson: … I think

toolkits/global/packages/global-forms/README.md
Heydon: @stuRobson this determines which type of field it is (text,email,radios,select etc). The verbose name is because some templates may need to be registered globally and there may already be something in the global scope simply called `text`.

Heydon: > does this need to be globalFormEmail or does it not matter?

Ah, I see what you mean. Yeah, that would be better.

sturobson: 👍🏼

toolkits/global/packages/global-forms/demo/context.json
sturobson: 👍🏼

toolkits/global/packages/global-forms/HISTORY.md
sturobson: 👍🏼

sturobson (approved):

PR #765, opened 2022-07-13
updates SVG ID to match other SVG IDs

sturobson:
## Description

When checking the [global-breadcrumbs](https://elements.public.springernature.app/nature/components/global-breadcrumbs) we noticed that the chevron icons were not being displayed. 

On investigation the `#id` begin used in the toolkits were different between what was referenced and what existed.

Also, spotted that Nature's Section Heading also had the erroneous naming convention.

## Work carried out:

- SVG `#id` change to remove unneeded `icon-`.
- Demos re-run for `global-breadcrumbs` and `nature-section-heading`.
- bumped the `brand-context` version for `global-breadcrumbs`.
- updated `package.json` for both components.
- updated `history.md` for both components.

notes: As these are more "spelling error" or naming convention bugs, I've gone with a `x.x.N` patch release for both components.

## Testing

As well as the test run by the `npm run demo` command, I also checked these locally (on a local live server) in Chrome, Safari, and Firefox.

#### Before

Screenshot 2022-07-13 at 11 49 56

#### After

Screenshot 2022-07-13 at 11 50 10

#### Before

Screenshot 2022-07-13 at 11 50 34

#### After

Screenshot 2022-07-13 at 11 50 38

## Other

This, however, does not fix the missing SVG icons in the Elements Design System site. As the `iframe`d HTML cannot go up the file path to use the SVG … this is a separate issue, not effecting this PR. 

A cursory mention of [this PR](https://github.com/springernature/frontend-toolkits/pull/719) that would deal with the file size of SVGs by running them (manually) through SVGOMG


Heydon (approved):
LGTM (Let’s Get This Merged)

PR #769, opened 2022-07-14
Section heading docs

amyhupe:
This PR:

    * assigns sensible default heading levels to Section heading variants with option to override
    * adds guidance on when to use variants with and without top border
    * adds guidance on overriding default heading levels
    * updates guidance on adding chevron icon to section headings as link

### DIFF
--- a/toolkits/nature/packages/nature-section-heading/README.md
+++ b/toolkits/nature/packages/nature-section-heading/README.md
@@ -4,7 +4,7 @@
 
 The Nature Section heading is a type of heading used for titles of some subsections of pages on Nature.com.
 
-It uses borders to the heading text to make it stand out from other headings on the page.
+It uses borders to make it stand out from other headings on the page.
 
 ## When to use this component
 
--- a/toolkits/nature/packages/nature-section-heading/README.md
+++ b/toolkits/nature/packages/nature-section-heading/README.md
@@ -20,11 +20,12 @@ Write clear, descriptive headings using sentence case.
 
 ### Section headings with and without a top border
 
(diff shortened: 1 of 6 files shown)


toolkits/nature/packages/nature-section-heading/README.md
```diff
@@ -4,7 +4,7 @@
 
 The Nature Section heading is a type of heading used for titles of some subsections of pages on Nature.com.
 
-It uses borders above and below the heading text to make it stand out from other headings on the page.
+It uses borders to the heading text to make it stand out from other headings on the page.
```
jon-stevens, line 7:
```suggestion
It uses borders to make it stand out from other headings on the page.
```

toolkits/nature/packages/nature-section-heading/view/section-heading.hbs
```diff
@@ -1,14 +1,24 @@
-<div class="c-section-heading {{no-bt}}">
-	{{#if url}}
-	<h2>
-		<a href="{{url}}" class="c-section-heading__link">
+<div class="c-section-heading {{#if no-bt}}{{no-bt}}{{/if}}">
```
jon-stevens, line 1: I think with handlebars you don't need to check the expression exists, if it doesn't it won't render anything.
Although I guess it makes it clear this way.

alexkilgour: I think it's better to be clear and explicit here. I'm not sure if that functionality is true for older versions of handlebars, and some people will be less familiar with the syntax.

PR #783, opened 2022-08-02
Opt in to line truncation and debug removal

Heydon:
```
## 6.0.0 (2022-08-02)
    * BREAKING:
        * line-clamping now _opt in_
        * Debug removed
```


amyhupe (approved):
Makes sense to me, thanks @Heydon

PR #791, opened 2022-08-11
Card hyphenation

Heydon:
```
## 8.3.1 (2022-08-11)
    * Activate hyphenation for German content
```

## Before

text overflows

## After

text hyphenates and wraps


benjclark (approved):
👍

PR #794, opened 2022-08-31
Forms rc.6

Heydon:
Changes to global forms and the complementary component `global-status-message`:

## Global forms:

```
## 5.0.0-rc.6 (2022-08-31)
    * `display: none` for inputs without labels (hidden inputs)
    * Larger checkboxes and radios
    * Correct grey for checkbox and radio borders
    * Bright blue focus style
    * `tabindex="-1"` for programmatic focus compatibility on error summary
    * Optionally exclude discrete errors in error summary
```

## Global Status Message:

```
## 4.0.0 (2022-08-31)
    * Message section now permits more than plain text
    * Message now focusable programmatically
    * Support for live regions
```


toolkits/global/packages/global-forms/view/globalFormErrorSummary.hbs
```diff
@@ -7,13 +7,15 @@
...
 	<div class="c-forms__error-summary-text">
-		<div class="c-forms__error-summary-title" id="{{id}}">{{title}}</div>
-		<ul class="c-forms__error-summary-errors">
-			{{#each errors}}
-				<li>
-					<a href="#{{id}}">{{error}}</a>
-				</li>
-			{{/each}}
-		</ul>
+		<div class="c-forms__error-summary-title" id="{{id}}" tabindex="-1">{{title}}</div>
```
morgaan, line 10: We actually took the path to have the `tabindex="-1"` onto the aside element.
Any particular reason you enforce it on the title rather?

On our side that let the focus ring surrounds the aside rather than the title itself.

foxintherain (approved):

toolkits/global/packages/global-status-message/README.md
```diff
@@ -12,114 +12,115 @@ The `global-status-message` component uses the `DEFAULT` branding across _all_ o
...
+<!-- first add the wrapper -->
+<div class="c-status-message" role="status" aria-live="polite">
+</div>
+
+<!-- then add the contents -->
+<div class="c-status-message" role="status" aria-live="polite">
+    <svg class="c-status-message__icon" width="24" height="24" role="img" aria-label="warning:" focusable="false">
+        <use xlink:href="path/to/warning.svg#i-warning"></use>
+    </svg>
+	<div class="c-status-message__message" tabindex="-1"{{#if id}} id="{{id}}-message"{{/if}}>
```
morgaan, line 116: I recognise the pattern from the error-summary in regards to tabindex. So I would raise the same comment. At least so far it is consistent.

toolkits/global/packages/global-status-message/README.md
```diff
@@ -12,114 +12,115 @@ The `global-status-message` component uses the `DEFAULT` branding across _all_ o
...
+		A new warning message
+	</div>
+</div>
+```
+
+If you are compiling the handlebars template in the client, this is possible by omitting the `message` and `iconURL` properties in the initial data.
```
morgaan, line 122: Sorry, not quite getting this, can you iterate to something clearer 😊

toolkits/global/packages/global-status-message/view/globalStatusMessage.hbs
```diff
@@ -0,0 +1,12 @@
+<div class="c-status-message{{#if boxed}} c-status-message--boxed{{/if}}{{#if type}} c-status-message--{{type}}{{/if}}"{{#if liveRegion}} role="status" aria-live="polite"{{/if}}>
```
morgaan, line 1: Shall we have a `tagContainer` prop that would care for using `aside` element for semantic and accessibility sake. FYI, here is our current implementation of the status message partial:
https://github.com/springernature/researcher-portal/blob/master/researcher-portal-frontend/src/view/partials/status-message.hbs

morgaan (commented):
This is looking awesome so far thank you for this. We dropped (@andreaskoscinski, @danielclubb and I) a few comment. Looking forward to your feedback. Cheers!

toolkits/global/packages/global-forms/view/globalFormErrorSummary.hbs
Heydon: @morgaan I put it there because you were have issues with announcement because it was not on a text element? You could use `:focus-within` to set the focus style on the container 🕶

toolkits/global/packages/global-status-message/view/globalStatusMessage.hbs
Heydon: Ah, I forgot to make it an aside. I don't think we need an option, I have now set it as an `<aside>` as standard.

toolkits/global/packages/global-status-message/README.md
Heydon: @morgaan I think it is clearer now? (no need to mention the client part, in fact):

> _You can achieve the initial state (without the contents) by omitting the `iconURL` and `message` properties in the data._

amyhupe:
Need to unmerge this

Issue #763, opened 2022-07-11
Design Tokens Work To Do


sturobson:
Now that we have published the initial generated `.scss` files from the Design Tokens `.json` files and introduced the idea of 'component tokens' with the newly refactored/rewritten `global-forms` component there have been a few things crop up that could be worked on to better serve the front-end toolkits with Design Tokens.

## Work to be carried out

- [x] add alias tokens for spacing (`gap`, `padding`, `margin`).
- [x] shorten `tokens--` namespacing to `t-` to match `u-`, `c-`, and `l-`.
- [x] fix misspelling of components for outputted file header.
- [x] remove springer tokens from global-forms.
- [x] prepare the `design-tokens` folder for npm
- [x] move dependencies to `design-tokens` folder, remove from `brand-context`.
- [x] change `npm scripts` to be more modular, and less likely to create constant breaking changes.
- [x] create an npm script that you pass arguments to to generate only one component `.scss` file? 
- [x] recompose the component tokens for `global-forms` now there's a better understanding on how they have been implemented.
- [x] move parts of the `readme.md` to `literal`, `alias`, `component` folders as needed.
- [x] add setting up and usage guidelines to the root `readme.md`.
- [x] create a `history.md` file to keep a history (in lieu of and in readiness of it being on `npm`).
- [x] add linear gradient (#0070A8 to #01324B) to Springer Nature Tokens for the Springer Nature Header background.
- [x] update line heights so the default is `md`.
- [x] change grayscale to greyscale to match Design documentation and usage.

---

There's a good chance I'll be adding to this list of things, but I hope to get all of this into one PR.

amyhupe:
Blocked until Stu's back from leave in September

PR #806, opened 2022-09-12
Fixes `packages/packages` error when running full tokens build.

sturobson:
Because there are currently no brand specific components that make use of design tokens I had an empty `''` in the `build-component.js` file which would create a `package/package` folder in toolkits. 

This is a hotfix becuase I'm working on a better tokens workflow with #763. 

I have removed the brand components task from being run when running `npm run tokens:component`.

I have commented out the erroneous code.


Heydon (approved):
Lend Gary That Mower

PR #792, opened 2022-08-27
Adds pictographic radio inputs to Global Forms

benjclark:
demo4

This PR adds support for pictographic radio elements.

- Supports inline SVGs only (keeping it simple for now)
- Adds settings for pictographic radio to enable visual appearance and accessibility to be managed effectively. These are all described in the README.
- Parts of the SVG styling have been set using both Sass variables and CSS variables as options. This allows for widths, heights, and state colours to be set per image.

**Try it out**
1. `npm run demo -- -p global-forms`
2. Open `dist/index.html`

PR #810, opened 2022-09-14
Sensible width and height for aria-hidden SVGs in core browser styles

benjclark:
Thanks to @morgaan  for raising an issue in https://github.com/springernature/frontend-toolkits/pull/792 which highlighted the need for us to be setting a sensible default width and height for all decorative inline SVGs in our core styles.

In some cases we use inline SVGs on our sites and style them in our enhanced stylesheets. Those enhanced styles often set an appropriate width and height for the inline SVG. It makes sense to have a sensible default in the core styles, this prevent the core browser user experience from looking a bit broken (because of huge SVGs that have scaled to the size of their container).

### DIFF
--- a/context/brand-context/HISTORY.md
+++ b/context/brand-context/HISTORY.md
@@ -1,4 +1,6 @@
 # History
+## 27.1.0 (2022-09-14)
+      * Sets a sensible default width and height for aria-hidden inline SVGs in the core styles to reduce the core experience looking broken
 ## 27.0.0 (2022-09-09)
     * BREAKING:
       * Springer Nature: Updates body font family to Merriweather Sans.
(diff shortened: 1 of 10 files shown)


sturobson:
If it's for sensible defaults, I would have thought it makes more sense that this would be in a (as yet uncreated) `basic.scss` file `default`?

morgaan:
> If it's for sensible defaults, I would have thought it makes more sense that this would be in a (as yet uncreated) `basic.scss` file `default`? One place to maintain, rather than 3.

Yes Stuart was quicker than me on that one 😊

benjclark:
@sturobson @morgaan Thanks both. Yes I considered this but it seemed to me that we have intentionally defined `basic.scss` per brand and not a shared default. Default only imports a normalize stylesheet and leaves the basic core styling up to the brand. @alexkilgour can you remember if there were any particular reasons behind that?

I take your point though. What we could do here is extract the common CSS declarations from that are the same in all 3 `basic.scss` files and created a `basic.scss` file in default. I can't see any harm in doing that.

I'll do that as part of this change then 👍

sturobson:
> What we could do here is extract the common CSS declarations that are the same in all 3 basic.scss files and created a basic.scss file in default. I can't see any harm in doing that.

PR #811, opened 2022-09-14
Universal spacing

Heydon:
```
## 28.0.0 (2022-09-14)
    * BREAKING:
        * New universal block spacing (inter-element margin) for flow content
    * block-spacing alias design tokens
    * spacing applied to heading utility classes
    * “accessible focus styles” removed (redundant `outline: none` on non-interactive elements)
```

## Comparison of brands
Screenshot of brand flow content spacing
Screenshot of brand flow content spacing
Screenshot of brand flow content spacing
Screenshot of brand flow content spacing

PR #812, opened 2022-09-15
updates brand context to version 28

sturobson:
introduces block spacing tokens and Sass variables … a precursor (chicken or egg, ¯\_(ツ)_/¯) to allow #811 to not make Travis cry. 

*note:* 

All this work was created by @Heydon, but as there is a problem when needing to update the brand context whilst working on components we cannot put both changes in one PR … because Travis pulls down the version of brand-context from npm … if it doesn't exist it fails.
To try to get #811 at a point where it can be merged in, I've pulled out the various brand-context changes and made this PR … because of how the Travis CI is currently set up).

### DIFF
--- a/context/brand-context/HISTORY.md
+++ b/context/brand-context/HISTORY.md
@@ -1,6 +1,12 @@
 # History
+## 28.0.0 (2022-09-14)
+    * BREAKING:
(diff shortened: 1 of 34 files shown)


Heydon (approved):
Leslie Grantham Blood Type

PR #817, opened 2022-09-26
Nature header bugfix

ricky-lau:
* Move subscribe link into menu list and remove redundant div container in template
* Remove margin right on menu to inline menu items sooner as viewport gets wider
* Update brand-context v28.0.1


toolkits/nature/packages/nature-header/package.json
```diff
@@ -1,9 +1,9 @@
 {
   "name": "@springernature/nature-header",
-  "version": "9.0.0",
+  "version": "9.0.1",
```
alexkilgour, line 3: This is a breaking change to the `brandContext` which means we need a breaking change here

alexkilgour (approved):
👌

PR #818, opened 2022-09-28
Updates border alias tokens and Pictographic Radio component tokens

benjclark:
Following discussion with James A in the Elements team. These changes need to be made.

### DIFF
--- a/context/brand-context/HISTORY.md
+++ b/context/brand-context/HISTORY.md
@@ -1,7 +1,7 @@
 # History
 ## 28.1.0 (2022-09-28)
     * Updates border alias design tokens to use grayscale.500 for input borders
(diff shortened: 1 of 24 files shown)


sturobson:
Currently, there's no automation in creating the relevant Sass files from changes in a tokens `.json` file (this could be improved if/when the design tokens are there own nom package. Also, there's no way to only generate component specific tokens as `sass` as needed (this is being worked on in #801). 

To work through this we could run the `nom` task to generate new tokens and only commit the relevant file changes associated with this PR. 

Happy to jump on this to do that.

benjclark:
Thanks @sturobson 

I have now generated settings from tokens for the settings files that live in Brand Context.

There is [a separate PR](https://github.com/springernature/frontend-toolkits/pull/818) that will generate settings from these token changes for components that will need to bump to this brand context version.

PR #821, opened 2022-09-30
Global forms 5 (full release)

Heydon:
```
## 5.0.0 (2022-09-29)
    * BREAKING:
        * Removal of `optional` property; can be handled as part of the label or hint instead
    * Support `disabled` property on select options
    * `requiredSuffix` property for adding an asterisk or equivalent
```


sturobson:
Oh yeah

toolkits/global/packages/global-forms/README.md
```diff
@@ -80,10 +75,10 @@ The `hint` property adds hint text under the main label text but _inside_ the `<
 "hint": "This will be the email address you used when registering"
 ```
 
-The `optional` property adds _“(optional)”_ to the label text.
+To accompany the `required` property, there is `requiredSuffix`. You can use this to add an indicator after the label text, such as an asterisk. HTML is permitted, meaning you can hide the asterisk from assistive technologies using `aria-hidden`. The presence of the `required` attribute on the input itself suffices for screen reader software.
```
amyhupe, line 78: As discussed on Slack, I think my preference would be to go back to the previous implementation where we have an `optional` property and have fields (implicitly) required by default.

This is because:

1. I hope it might discourage the collection of too much non-essential user data - which is general good practice
2. I think appending (optional) to a field's label is clearer than using an asterisk to denote a required field (though I don't have evidence for this and I accept that the latter is a common pattern).

After discussing with Head of Content Design - we think that appending (optional) to the field's label would be the clearest implementation.

toolkits/global/packages/global-forms/README.md
```diff
@@ -80,10 +75,10 @@ The `hint` property adds hint text under the main label text but _inside_ the `<
...
 
 ```json
-"optional": true
+"requiredSuffix": "<span aria-hidden=\"true\">*</span>"
```
amyhupe, line 81: See previous comment. 

I also think it might be good to add some guidance around this, e.g.:

Avoid asking for information you don't really need. If you do need to ask for optional information, set the `optional` property to `true`.

toolkits/global/packages/global-forms/README.md
```diff
@@ -1,7 +1,5 @@
 # Global Forms
 
-**IMPORTANT:** _This is a **release candidate** representing an ambitious reworking of the global-forms component, including design tokens integration, at a component level, for the first time. It is expected to need some additional work before a full release can be made. Tokens are generated to the `scss/00-tokens` folder and cannot be edited directly. If you need to temporarily add or override variables, please do this in a `10-settings` folder, one level down. The Elements Design System team will assess these changes for making these variables tokens later._
-
 This component includes a number of form fields and related templates. It is designed to make it as simple as possible to create an HTML form.
```
amyhupe, line 3:
```suggestion
This component includes a number of form fields and related templates.
```
Second sentence feels kind of redundant to me, on reflection

amyhupe (commented):
Looks good to me aside from the couple of comments I've added.

I would like to add some design guidelines in the near future, but let's keep things simple for now.

Heydon:
> As discussed on Slack, I think my preference would be to go back to the previous implementation where we have an optional property and have fields (implicitly) required by default.

@amyhupe To be clear, fields are _not_ currently required by default. But I think they should be. At the moment, both `required` and `optional` properties are opt-in and not contingent on one another.

Here's what I propose:

1. All fields are `required` by default
2. Setting `optional: true` makes them not required...
3. ...and appends ”(optional)” to the label, as discussed with Gaynor
4. provide opt-in `optionalSuffix` so the "(optional)" text can be overridden

We can leave (4) undocumented, at least for now; just an escape hatch.

The above would be significant breaking changes. Since SPA team needs the other change (`disabled` for `<option>` elements) ASAP, I suggest I separate these and do an rc10 for this urgent change first, followed by a full release with optional-by-default.

(I hope using "optional", "opt-in", and `<option>` all in the above missive does not make it completely indecipherable. I chose "opt-in" so I wqsn't saying "make required optional" or "make optional optional" etc etc 🥴)

amyhupe:
@Heydon 

I'm happy with this Heydon!

toolkits/global/packages/global-forms/README.md
Heydon: @amyhupe I removed `requiredSuffix` now, since this is redundant in a required-by-default approach.

toolkits/global/packages/global-forms/README.md
Heydon: @amyhupe See above. Please check the rewording of that section.

toolkits/global/packages/global-forms/README.md
```diff
@@ -74,17 +69,23 @@ This component supports a wide range of standard form field attributes. For exam
...
 "hint": "This will be the email address you used when registering"
 ```
 
-The `optional` property adds _“(optional)”_ to the label text.
+#### Required
 
-```json
-"optional": true
-```
+Fields are required by default, because you should not be asking for unnecessary data.
```
amyhupe, line 82:
```suggestion
Fields are required by default. Avoid asking for information you don't really need. 
```

toolkits/global/packages/global-forms/README.md
```diff
@@ -74,17 +69,23 @@ This component supports a wide range of standard form field attributes. For exam
...
 
-The `optional` property adds _“(optional)”_ to the label text.
+#### Required
 
-```json
-"optional": true
-```
+Fields are required by default, because you should not be asking for unnecessary data. 
+
+The `required` and `aria-required="true"` attribution is applied automatically. Unless `novalidate: true` is applied on the field’s data, browser validation will warn the user when they have not filled out the field upon submission.
```
amyhupe, line 84: Should this say "attribute" rather than "attribution"?

toolkits/global/packages/global-forms/README.md
```diff
@@ -74,17 +69,23 @@ This component supports a wide range of standard form field attributes. For exam
...
+
+In the rare cases a field _is_ optional, it can be marked as such using `optional: true`. This will remove `required` and `aria-required="true"` as well as appending the text “(optional)” to the field label. This is the agreed way to communicate a field can be skipped.
```
amyhupe, line 86: Just trimmed this down a bit. Suggest removing the last sentence as, *in theory* everything we do in Elements should be the agreed way to do it.

```suggestion
If you do need to ask for optional information, set `optional: true`. This will remove `required` and `aria-required="true"` as well as appending the text “(optional)” to the field label. 
```

amyhupe (commented):
A couple of small changes from me but on the whole it looks good!

toolkits/global/packages/global-forms/README.md
Heydon: @amyhupe No, I'm listing two attributes.

Heydon: "attributes are applied automatically" is probably simpler, though

sturobson (dismissed):
Louis Goes To Market

sturobson (approved):
Lexus Grand Tourer Motorcar

PR #823, opened 2022-09-30
Adds Global Ratings Survey component

benjclark:
### Aim
The aim of this component is to gather customer feedback using a low friction approach (choose a rating 1 to 5; awful to great) to maximise the amount of responses we gather. The intention is to deploy this component on many of our user journeys across many of our websites.

The solution is comprised of:
- this Global Ratings Survey component for the frontend
- Google Tag Manager and Google Cloud: Big Query for the backend
- Survey Monkey for any qualitative survey questions we wish to use in addition to the standard rating gathering
- Data Studio for the reporting

I have created a walkthrough video for this component which I hope makes it easier to understand it in the context of the overall solution:
https://drive.google.com/file/d/1DwatHb9ugJFkhMsdJCKGq87rriy68WWH/view?usp=sharing

### What this PR does
- Adds version 0.0.1 of this component, i.e. the component will not be consumable when this is merged (will become 1.0.0 once *all* documentation finished and planned initial in-situ testing phase completed)
- Adds the technical usage documentation to the README but does not define the other aspects (this will be done in due course but is considered not part of this PR). A TODO for other aspects are defined in initial code block in README file.
- Adds a demo page that demonstrates the layouts this component is designed for

#### JS dependent
This is a js dependent component. If js is disabled then the component is not visible. This was deemed acceptable given the function of this component (relating purely to analytics gathering and dependent on GTM).
#### Can be used if application doesn't use Elements
Not all of our applications are using Elements. Therefore instructions have been provided for how to use this component if you do not use Elements (via generating demo.html and copying). Consequently I have hardcoded the icon used in the component's success message instead of using the `<use xlink:href=""></use>` mechanism.
#### Accessibility
The component contains visually hidden text to improve its accessibility and has been tested by me using VoiceOver and NVDA. Any feedback relating to its accessibility welcomed. If others could test the accessibility of it as well I would really appreciate it.
#### Hard coded button styles
At the time of writing button styling in Elements is in need of attention and refactor, consequently there is not one "type" of button available that would be satisfactory for all imprints. Given the intention for this component to be used across many of our websites I have hardcoded the button styles for now. These should be swapped over to Brand Context button styles once they have been improved.
#### Absolute design token values used
Some of our websites use `html {font-size: 62.5%}`. Consequently this component uses absolute design token values as well as some hard coded values. This is to guarantee a consistent look and feel on all our websites. See [this discussion](https://springernature.slack.com/archives/CNBTFLBLP/p1665752342597439?thread_ts=1665745038.961489&cid=CNBTFLBLP) on Slack

localhost_63342_frontend-toolkits_toolkits_global_packages_global-ratings-survey_demo_dist_index html__ijt=3rl09oq5rcnvu784i3l37c5dd8 _ij_reload=RELOAD_ON_SAVE

localhost_63342_frontend-toolkits_toolkits_global_packages_global-ratings-survey_demo_dist_index html__ijt=3rl09oq5rcnvu784i3l37c5dd8 _ij_reload=RELOAD_ON_SAVE (2)

localhost_63342_frontend-toolkits_toolkits_global_packages_global-ratings-survey_demo_dist_index html__ijt=3rl09oq5rcnvu784i3l37c5dd8 _ij_reload=RELOAD_ON_SAVE (3)

localhost_63342_frontend-toolkits_toolkits_global_packages_global-ratings-survey_demo_dist_index html__ijt=g0klil3pnadcr8831sj5ck8785 _ij_reload=RELOAD_ON_SAVE


sturobson:
re: Design Tokens … 

Looking good.

The ideal would be to follow this (current) suggested [naming convention](https://github.com/springernature/frontend-toolkits/blob/b5c0820bced600cfbceb959f5b7c43f9a5439a80/context/design-tokens/README.md#component-tokens)

<component>-<property>-<category>-<priority>-<type>-<state>-<validation>-<value>

> Where:
> 
> <component> is the component of the token (eg: button).
> <property> is the property of the token (eg: color).
> <category> is the category of the token (eg: spacing).
> <priority> is the priority of the token (eg: primary).
> <type> is the type of the token (eg: solid).
> <state> is the state of the token (eg: hover).
> <validation> is the validation of the token (eg: valid).
> <value> is the value of the token (eg: 2rem).

There's also some improvement in the [Design Tokens Schema](https://github.com/springernature/frontend-toolkits/blob/b5c0820bced600cfbceb959f5b7c43f9a5439a80/context/design-tokens/README.md#design-tokens-schema)

## Example
```json
{
    "aspect-ratio": {
        "1x1": {
            "name": "1x1",
            "value": "1",
            "description": "This is to make an item square",
            "type": "ratio",
            "category": "aspect-ratio",
            "themeable": false,
            "meta": {
                "public": true,
                "deprecated": false,
                "documented": true,
                "experimental": false,
                "SassVariable": true,
                "CSSCustomProperty": false
            }
        }
    }
} 
```

I think the schema can be updated in the future (the idea is we can incorporate this information in a components Documentation page, which requires the extra key/value pairs). If we could try and follow the naming convention a little more, that'd be great. Happy to jump on a call and suggest things -- or do some inline comments.

note: the above is in [a separate PR](https://github.com/springernature/frontend-toolkits/pull/835)

toolkits/global/packages/global-ratings-survey/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "@springernature/global-ratings-survey",
+  "version": "0.0.1",
+  "license": "MIT",
+  "description": "html form based component for gathering user rating data",
+  "keywords": [],
+  "author": "Springer Nature",
+  "brandContext": "^29.0.3",
+  "dependencies": {
+    "@springernature/global-forms": "^5.0.0"
```
sturobson, line 10: As a single installable thing I'm guessing because we are using the `global-forms` as a dependency here we get the `brand-context` by default, because it's a dependency of the `global-forms`?

benjclark: Yes good point. And consequently there is a possibility that at any time this component effectively has a dependency on two different versions of Brand Context at the same time 😬 (this package.json and Global Form's package.json)

Removing the dependency on Global Forms would result in the entire pictographic radios implementation being duplicated in to this package though. Which seems very unappealing.

Hmm 🤔

sturobson: The `demo/index.html` is using brand-context stuff -- so I think we can leave this as it is. 

Personally, I think it's better to have components consume other components than duplicating things because we cannot do that. 

hashtag resolved.

toolkits/global/packages/global-ratings-survey/scss/50-components/_ratings-survey.scss
```diff
@@ -1,10 +1,17 @@
 .c-ratings-survey {
 	padding: $global-ratings-survey-container-padding;
+	margin: 0 (0 - $global-ratings-survey-container-padding);
```
Heydon, line 3: @benjclark  Could you explain why a negative margin was the solution here? Nothing wrong with negative margins _per se_ but I'm wondering why you needed to do this here.

Also, I'm a bit confused by the syntax. I think I'd write it as

```
margin: 0 calc(0 - #{$global-ratings-survey-container-padding});
```

But then again, I'm not the biggest Sass person.

toolkits/global/packages/global-ratings-survey/README.md
```diff
@@ -0,0 +1,134 @@
...
+- Designed to be unobtrusive - can be placed any page at any point in a user journey
+## When to use this component
+- Use this component for the CSAT method
+- Used in combination with other survey tools
+  - Could be used in combination with an additional survey at the end of a task / exit survey
+## When not to use this component
+- Do not use this component for NPS. NPS  is a different use-case - not an either or. If you need to use NPS - it should be placed at the end / exit of a task.
+## How to use (setup)
+Who to contact
+The question can be not changed
```
Heydon, line 26: @benjclark Are these your own notes?

toolkits/global/packages/global-ratings-survey/scss/50-components/_ratings-survey.scss
```diff
@@ -10,7 +10,11 @@
 	justify-content: center;
 }
 
-.c-ratings-survey__heading {
+// Specificity score increased to ensure overrides heading styles on all of our imprint sites
+.c-ratings-survey h1.c-ratings-survey__heading,
```
Heydon, line 21: @benjclark I'm thinking `!important` is better than making the selector somewhat more complex?

Heydon (commented):
I left a couple of comments, nothing major. The one about the README containing your own notes seems to be from an earlier commit. But the README still seems to contain a private comment about the document structure, at the top? Not sure if this will be suppressed in the output, like a comment? That's the only thing I think definitely would need addressing before approval.

toolkits/global/packages/global-ratings-survey/package.json
alexkilgour, line 10: I see this conversation has already been marked as resolved, but one of the underlying principles of the design system, as it is now, is that components CANNOT depend on other components. If/when we move to a single release then inter-component dependencies is not so much of an issue and we can rethink the guidance.

It is almost inevitable in this scenario that you will end up with the version of context defined for this component and the version defined for (this version of) global-forms becoming out of sync.

> there is a possibility that at any time this component effectively has a dependency on two different versions of Brand Context at the same time

This is not the case,  we are only ever dependent on _the version defined in this file_ (`29.0.3`) - even by using that version of global forms the context that is installed with this component will be the one used. We just have to hope that forms works with the one used. This is how it works for the demon, for a consumer you are always reliant on a compatible version of context being installed directly from their dependencies (which I am not always sure is the case, but should be an issue solved by the single version)

In summary; I don't see an option of doing anything other than what you have done in order for this to work properly, but at the moment it is most definitely an anti-pattern and not something we want done anywhere else. As such I think (1) we need to get the single version work done as a priority, (2) this component should never be any more than experimental (`0.0.x`) until that is done, and (3) We should provide a very clear on the README that this is temporary until the above happens, and is NOT a pattern to be followed at this time.

toolkits/global/packages/global-ratings-survey/README.md
alexkilgour: Yep, looking at the rendered file I don't think any of this text should be appearing?

toolkits/global/packages/global-ratings-survey/README.md
```diff
@@ -0,0 +1,134 @@
...
+Install the Global Ratings Survey package:
+```shell
+npm install @springernature/global-ratings-survey
+```
+You will also need to ensure you have its dependencies installed. These are Brand Context and Global Forms:
+```shell
+npm install @springernature/brand-context
+```
+```shell
+npm install @springernature/global-forms
```
alexkilgour, line 46: All these instructions will install the latest versions of all these components which is NOT what you want to be doing. You need to tailor these to specify the exact version of each of these e.g. `npm install @springernature/global-ratings-survey@0.0.1`

This is probably the place to add the caveat about this being an anti-pattern that I mentioned in a previous comment

toolkits/global/packages/global-ratings-survey/README.md
benjclark: Sorry, I have removed this now. They were there as James A and Niamh had made TODO notes on the documentation and this seemed like the best place to store them at the time. I have now created a documentation site for this overall solution: https://customer-satisfaction-survey.public.springernature.app/docs/introduction and so I will store the TODOs as GitHub issues against that repo instead

toolkits/global/packages/global-ratings-survey/scss/50-components/_ratings-survey.scss
benjclark: Thanks Heydon. 

I was using negative margin to eek out as much horizontal space as possible to allow for nicer amounts of spacing between icons. 

But looking at this again with fresh eyes I think it's a bad idea in general that would have some overflow issues when the component is placed in a location where there isn't any page section padding to overflow into. So I'm scrapping the idea entirely and will just reduce the gap between icons a little bit to ensure the design is ok and is not wrapping on devices such as iPhone 5 that have only 320px vw to play with.

toolkits/global/packages/global-ratings-survey/package.json
benjclark: I have added warnings to the README for points (2) and (3). Both of those warnings can be removed after (1) is done, as discussed.

toolkits/global/packages/global-ratings-survey/README.md
benjclark: Good point, thanks!

toolkits/global/packages/global-ratings-survey/scss/50-components/_ratings-survey.scss
```diff
@@ -0,0 +1,117 @@
...
+.c-ratings-survey a {
+	display: inline-block;
+	margin-top: $global-ratings-survey-link-margin;
+	font-family: inherit;
+	font-size: $global-ratings-survey-message-font-size;
+}
+
+.c-ratings-survey__visually-hidden,
+.c-ratings-survey legend.c-forms__label {
+	/* hide visually, not to assistive tech */
```
alexkilgour, line 55: does this do something different to using the [visually hide](https://github.com/springernature/frontend-toolkits/blob/main/context/brand-context/default/scss/30-mixins/hiding.scss#L16) mixin available in `brand-context`?

available by using `@include u-visually-hidden`

alexkilgour: Oooo, good catch. In which case this looks good, and yeah if you could add as an issue that would be wonderful 😃

sturobson (approved):
LGTM - let's go!

PR #826, opened 2022-10-05
Support for dashed lists

moddular:


sturobson:
I wonder if this needs a little more work and maybe design sign-off?

The placement of the `-` looks out of place compared to the available types of styles, and I'd hazard a guess that there's the chance it might be slightly different depending on which front type face is being used too.

Screenshot 2022-10-05 at 12 07 01

moddular:
Hmm, odd the default display of a string marker lines up so poorly with other types.

Tweaking font-size/line-height on ::marker gets somewhat closer

image

sturobson:
Typography design – 

> "Hyphens are customarily positioned slightly above center of a font’s x-height. This works well for running copy as well as oldstyle figures. But when used with (or next to) cap-height glyphs or lining figures, they appear too low."

https://creativepro.com/glyph-positioning-baseline-shift/

sturobson:
the only thing we have now is [the support for `::marker` and Safari](https://developer.mozilla.org/en-US/docs/Web/CSS/::marker#browser_compatibility) isn't as whole as Gecko and Chromium based browsers, but I think this can still proceed. 

I've tested in VoiceOver, by default you would get the bullet, square, disc announced before the list item text you don't when using `::marker`. You still get 'n of N' announced.

sturobson (approved):

PR #828, opened 2022-10-11
Removes default SVG sizing from core styles

benjclark:
The introduction of this style had oversights:

1. It has a higher specificity score than classnames because it is two selectors (an element selector and an attribute selector)
2. Having any default SVG styles for sizing means that html attributes cannot be used to size SVGs because the style always overrides the width and height html attributes

Therefore I think it should be removed


moddular (approved):
Thanks - looks good to me :+1:

PR #829, opened 2022-10-12
Add a utility class to reset margin-block-start

moddular:
Since upgrading to a version of brand-context which includes the block spacing styles we (pandora/orion/ecommerce) have been finding a few places where `margin-block-start` is creating unwanted spaces (before page headers, inside buttons etc...).

Given the specificity of the selectors for the block spacing it can't simply be overridden from within a class. This change provides a utility to brute force it back to 0 for places where it's needed.

### DIFF
--- a/context/brand-context/HISTORY.md
+++ b/context/brand-context/HISTORY.md
@@ -1,5 +1,8 @@
 # History
 
+## 29.0.2 (2022-10-12)
+    * PATCH: Add a utility class for resetting `margin-block-start`  
+
 ## 29.0.1 (2022-10-11)
     * PATCH: Removes default SVG sizing from core styles as it was overriding classname styles for icons and html width and height attributes
 
--- a/context/brand-context/default/scss/60-utilities/typography.scss
+++ b/context/brand-context/default/scss/60-utilities/typography.scss
@@ -100,3 +100,7 @@
 .u-h1+* {
 	margin-block-start: $tokens--typography-block-spacing-x-large;
 }
+
+.u-mbs-0 {
+	margin-block-start: 0 !important;
+}
--- a/context/brand-context/package.json
+++ b/context/brand-context/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@springernature/brand-context",
-  "version": "29.0.1",
+  "version": "29.0.2",
   "license": "MIT",
   "description": "Bootstrapping for all components and products",
   "keywords": [],


benjclark:
Yes, to echo Andrew, we have also found that the block spacing change has introduced unexpected spacing in many places across Nature, Link, BMC. I'm wondering if the CSS selector used in the block spacing styling might be a little too "loose" i.e. it selects too much and needs to be tightened up a bit. Particularly the wildcard used to selector any sibling. However, I'm not sure of what to propose in it's place.

What do you think @Heydon  @sturobson ?

benjclark:
image

This is an example, where we use an `<aside>` for advert content and the site header is adjacent to it - that div is subject to this block spacing style. It's feels a bit icky to have to apply a reset to `margin-block-start` in places like this, just to undo the block spacing style?

In the short term though maybe it's good to get this utility in to Elements so everyone has a utility to use to fix the issue for now?

benjclark (approved):

Heydon:
@moddular Thanks for this, Andrew, and apologies for the regressions. It's hard to make such far reaching changes without some fallout. Don't fret the `!important`, all utilities should have it really anyway, at least until cascade layers are better supported.

tulipdexter:
@Heydon This also makes an assumption that no classes will have `-c` in them except components.
Including, say `epic-thing`, or in my case elements that use `cc-*`, which is the cookie consent prefix.
The `[class*=c-]` perhaps should be `[class^=c-]`?

sturobson:
(aside: perhaps this needs a little tidy up: https://github.com/springernature/frontend-playbook/blob/main/css/bem-css.md) to avoid confusion or disparate naming conventions

tulipdexter:
Yeah that's a good point because I'm not sure that `c-` applies to _all_ code, e.g. stuff outside of elements.

Heydon:
> @Heydon This also makes an assumption that no classes will have c- in them except components.
Including, say epic-thing, or in my case elements that use cc-*, which is the cookie consent prefix.
The [class*=c-] perhaps should be [class^=c-]?

@tulipdexter  Good thinking. I used `*=` in case the `c-` class was not at the beginning. Which still may be the case. But I thought we were using data attrs for the `cc-` stuff?

tulipdexter:
@Heydon we only use data attributes for the JavaScript hooks, not the styling.
All styles are hooked off of a `cc-` prefix.
Shall I create a patch to fix that now, or are you planning a re-think?

tulipdexter:
This: https://github.com/springernature/frontend-toolkits/pull/830/files

Heydon:
@tulipdexter It's not much better than what I had (the component class may not come first), but put it in now if you need the fix ASAP 🙏 I will revisit the whole thing soon.

tulipdexter:
@Heydon this is a very temporary thing just to stop my styles going doolally while you are all working on the _real_ solution

Heydon:
@tulipdexter Understood, sorry for the bother.

sturobson:
I know folks say it's where overrides go to die - but what about creating a [`shame.scss`](https://csswizardry.com/2013/04/shame-css/) file and adding the temporary fix in there?

PR #830, opened 2022-10-12
Default brand context: tighten block spacing selector for c-

tulipdexter:
Ensures that the `margin-block-start` is only applied to classnames _starting with_ `c-` not _containing_ `c-`.


alexkilgour (approved):
👌

PR #831, opened 2022-10-12
Temporarily fix the chevron size to override the value coming from th…

moddular:
…is version of brand-context


benjclark:
image


Am I right in thinking you need `!important` ?

benjclark:
Oh there we go, you beat me to it with a better solution 😄

benjclark (approved):

moddular:
> Oh there we go, you beat me to it with a better solution 😄

"Better" 👀

benjclark:
😂 

Less disgusting?

PR #834, opened 2022-10-13
supports default branch rename

jpw:
The default branch name reflects an unsympathetic use of language and is not in line with our recommendations.

I'm happy to do the default branch rename just before merging this PR, but are there other systems outside the scope of this repo which might break as and when we rename? @alexkilgour

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ cache:
 # Restrict builds on branches
 branches:
   only:
-    - master
+    - main
     - /^\d+\.\d+\.\d+$/
 
 # build notification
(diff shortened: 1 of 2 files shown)


sturobson:
is it `default` or is it `main` … 

the travis file says `main` but the contributing.md file says `default`

jpw:
`main`. I am suggesting amending the docu to say "the default branch" and thereby no longer mentioning any branch name.

The previous version of the docu should have probably said `master` not "master" 😉

alexkilgour:
How do we handle existing branches that have been branched from master?

alexkilgour:
Other than travis though, I don't think that there is any blocker on making this change

sturobson:
> How do we handle existing branches that have been branched from master?

I _think_ GitHub aliases it somehow‽ We moved from `master` to `main` whilst I was working on EMBL's Design System and I'm pretty sure it 'just worked'.

Clearly, anecdotal evidence should be tried and tested rather than trusted here :sob: 

But it looks like [that is what they've done‽](https://github.com/github/renaming#whats-already-changed)

alexkilgour:
> But it looks like [that is what they've done‽](https://github.com/github/renaming#whats-already-changed)

Yep, that looks like we are covered, thanks!

PR #836, opened 2022-10-19
Block spacing containment

Heydon:
```
## 29.0.4 (2022-10-18)
    * Reduce scope of block spacing
```

This PR is to address the regressions created by trying to introduce universal block-spacing. Regrettably, some late changes were hastily added as part of my effort to complete the frontend portion of the new springer.com (a hard deadline).

I've spent some time rethinking the selectors, and included comments. They no longer include anything of high enough specificity to stop utility classes working. The "reach" of the selectors is also more limited. 

These will benefit from local testing with your respective applications, @tulipdexter , @moddular, @benjclark. Thanks for letting me know the bad news and for implementing the temporary solutions to give me time to rethink the algorithm. 

@alexkilgour has offered to add documentation regarding IE11 and logical properties.

@moddular I've left your utility reset class in there for the time being, remove it if you see fit.

As a side note, we should consider adding `!important` to utility classes, or recommending importing them as their own [cascade layer](https://css-tricks.com/css-cascade-layers/). While the block spacing no longer introduces specificity issues, we may encounter them elsewhere.


sturobson:
Thanks for this rework @Heydon <3 

> As a side note, we should consider adding !important to utility classes.

This can be automated [using the Design Tokens and Style Dictionary](https://github.com/springernature/frontend-toolkits/pull/736) if we use it to make up some of the utility classes.

Heydon:
>This can be automated https://github.com/springernature/frontend-toolkits/pull/736 if we use it to make up some of the utility classes.

Oh, neat!

context/brand-context/default/scss/40-base/normalize.scss
```diff
@@ -39,7 +39,6 @@ main {
 
 h1 {
 	font-size: 2em;
-	margin: 0.67em 0;
```
alexkilgour, line 42: As this is a file not generated by us there is a danger that if we make these changes here that they get overridden by any updates to the normalize code (we updated this fairly recently)

This normalize file also gets used for the `core` level whereas the block spacing does not so we are removing this default spacing from that core experience.

I know it is clunkier but I would be inclined to leave normalize as it is and then reset the margins _before_ including the block spacing file in the `enhanced` view with a message along the lines of "this resets the default margins from the core experience to allow them to be handled by the block spacing for enhanced view"

alexkilgour: > suggested moving to normalise.scss
We did look at it in the past. Is there are official sass version?

benjclark: Makes sense. Now committed.

alexkilgour (approved):
This looks good to test out now

PR #839, opened 2022-10-26
Fix popup border color to be accessible

anasanzsnd:
After a conversation with [Carlos Mucharaz in the #accessiblity channel](https://springernature.slack.com/archives/C9MS96R7T/p1665568412838009), we noticed that the border to the popup component is not accessible, having the minimum contrast ration less than 3:1. 
This PR changes the border color from `rgba(151, 191, 216, .3)` to use the token color `$t-border-color-primary`, a [core color for border](https://elements.springernature.com/springernature/styleguide/color#borders) that is compatible with all brands.

### DIFF
--- a/toolkits/global/packages/global-popup/HISTORY.md
+++ b/toolkits/global/packages/global-popup/HISTORY.md
@@ -1,5 +1,8 @@
 # History
 
+## 5.1.1 (2022-10-25)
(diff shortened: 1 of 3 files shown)


sturobson (approved):
LGTM - great to see a Design Token being used ❤️

PR #843, opened 2022-11-07
Add referrer to global ratings survey

dsmr:
Adds a method to the Global Ratings Survey class to allow us to append the user's window.location.href to Survey Monkey links as a custom variable

### DIFF
--- a/toolkits/global/packages/global-ratings-survey/view/ratingsSurvey.hbs
+++ b/toolkits/global/packages/global-ratings-survey/view/ratingsSurvey.hbs
@@ -15,7 +15,7 @@
 				</div>
 				{{#if questionUrl}}
 					{{#if questionText}}
-						<a href="{{questionUrl}}" target="_blank">{{questionText}} (opens in a new tab)</a>
+						<a href="{{questionUrl}}" data-ratings-survey="survey-link" target="_blank">{{questionText}} (opens in a new tab)</a>
 					{{/if}}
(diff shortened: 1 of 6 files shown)


benjclark:
Diana and I paired on this 🙂

toolkits/global/packages/global-ratings-survey/view/ratingsSurvey.hbs
```diff
@@ -15,7 +15,7 @@
 				</div>
 				{{#if questionUrl}}
 					{{#if questionText}}
-						<a href="{{questionUrl}}?id={{id}}" target="_blank">{{questionText}} (opens in a new tab)</a>
+						<a href="{{questionUrl}}" data-ratings-survey="survey-link" target="_blank">{{questionText}} (opens in a new tab)</a>
```
jpw, line 18: Links to external sites must have `rel="noopener"` [added for security reasons](https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/noopener) even though [browser support for implied `noopener` behaviour if opening links in new tabs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#browser_compatibility) is pretty good.

jpw (changes requested):
A few things to be looked-at please, thanks 🙂

jpw (approved):

jpw:
Good stuff, thank you 🚀

PR #844, opened 2022-11-14
fixes a naming issue and missing value issue

sturobson:
at some point block spacing tokens got moved around in the json and were no longer being compiled.

This fixes that.


Heydon (approved):
Let's Go To Malta (LGTM)

PR #845, opened 2022-11-15
fixes issue of naming

sturobson:
trying to reinstate the block spacing tokens led to a difference in naming conventions from compiled tokens in Sass.

They should be `t-` and in `40-base/block-spacing.scss` it was set as `t--` 

This fixes that.


andreaskoscinski (approved):
LGTM, pretty straight forward

PR #846, opened 2022-11-15
the one that fixes the `t-` `tokens--` naming in utility classes…

sturobson:
and I think this issue should finally be set to bed.


andreaskoscinski (approved):
LGTM, hope that's the last renaming missing.

PR #870, opened 2023-02-23
Improvements to documentation

alexhallspringer:
Updated documentation to follow new pattern, removed broken screenshots.

### DIFF
--- a/toolkits/global/packages/global-status-message/README.md
+++ b/toolkits/global/packages/global-status-message/README.md
@@ -111,7 +111,6 @@ Supply an `id` in the data to make the status message focusable programmatically
 
 Focusing the message ensures it is visible in the viewport and a screen reader will announce it. If it is not appropriate to move focus to the message, rely on the live region for announcement in screen readers.
 
-The status message will likely be the best focus position for continued keyboard interaction.
 
 ## Template
 
--- a/toolkits/global/packages/global-status-message/README.md
+++ b/toolkits/global/packages/global-status-message/README.md
@@ -1,8 +1,33 @@
 # Global status message
 
-## Usage
+Status message lets you add different types of message banners to a page. These can either be always on the page or appear based on certain conditions or events.
 
-First, include the necessary Sass files in your project. The `global-status-message` component uses the `DEFAULT` branding across _all_ of our products.
+## When to use this component
+
+Use a status message when you need to tell the user something that is additional to the main content of the page. 
+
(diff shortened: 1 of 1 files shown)


toolkits/global/packages/global-status-message/README.md
```diff
@@ -119,8 +101,22 @@ To trigger announcement of a new status message, the live region must be added t
 </div>
 ```
 
-You can achieve the initial state (without the contents) by omitting the `iconURL` and `message` properties in the data. 
+#### Status changed
+
+If the status message changes, update the `message` and `type` properties. This means the screen reader will announce the new message.
```
Heydon, line 106:
```suggestion
If you change the message after the page has loaded, a screen reader would announce the new message.
```

(It's not expected the original data for the message will change in the browser, only on the server. Changes are more likely to happen with JavaScript.)

toolkits/global/packages/global-status-message/README.md
```diff
@@ -119,8 +101,22 @@ To trigger announcement of a new status message, the live region must be added t
...
+#### Status changed
+
+If the status message changes, update the `message` and `type` properties. This means the screen reader will announce the new message. 
 
 ### Focus
 
-Where an `id` is supplied in the data, the status message can be focused programmatically. Focusing the message ensures it is visible in the viewport _and_ announces it in screen reader software. The status message is likely to also be the optimal focus position for (continued) keyboard interaction.
+Supply an `id` in the data to make the status message focusable programmatically.
+
+Focusing the message ensures that it is visible in the viewport and a screen reader will announce it.
```
Heydon, line 112: I don't think it's clear, on reflection, when to focus and when to just rely on the live region. 

```suggestion
Focusing the message ensures it is visible in the viewport and a screen reader will announce it. If it is not appropriate to move focus to the message, rely on the live region for announcement in screen readers.
```

toolkits/global/packages/global-status-message/README.md
```diff
@@ -119,8 +101,22 @@ To trigger announcement of a new status message, the live region must be added t
...
+
+The status message will likely be the best focus position for continued keyboard interaction.
```
Heydon, line 114:
```suggestion
```

Heydon (changes requested):
Couple of changes. Only suggestions, though, you may want to tweak them.

Heydon (approved):

PR #874, opened 2023-03-07
Updates and improves skip link documentation

alexhallspringer:


toolkits/global/packages/global-skip-link/README.md
```diff
@@ -17,40 +30,53 @@ To include `global-skip-link` in your application, you need to choose **ONE** br
...
-* When clicking anywhere in the page, focus will return back to the top
+### Focus
+
+The target section must be focusable so that users can then continue to tab through content. Most browsers will do this automatically. However, Internet Explorer 11 does not.
+
+To help skip link work for keyboard and assistive technology users in IE11, add a negative `tabindex` to the main element. 
+
+```html
+<main id="main" tabindex="-1">
+```
```
Sreilys, line 61: Do we need a little explanation on the tabindex? Something like: "<main> is not a control or a interactive content, so it does not accept focus, but adding tabindex="-1" allows it to accept focus"

Sreilys (approved):

toolkits/global/packages/global-skip-link/README.md
alexhallspringer: Good point. Something to keep in mind for the next iteration I think.

PR #877, opened 2023-05-04
Update README.md

sky-jack:
I wanted to propose an extra line to this documentation, as being fairly new to CSS custom properties it took me a bit of time to realise how to best to adjust these values using media queries and though useful for the purpose demonstration, adding the properties inline seems to make this difficult.

### DIFF
--- a/toolkits/global/packages/global-layout-with-sidebar/README.md
+++ b/toolkits/global/packages/global-layout-with-sidebar/README.md
@@ -51,7 +51,7 @@ Somewhat unconventional, but the component can take it!
 
 ### Custom settings
 
-The `global-layout-with-sidebar` component’s settings are mapped to CSS custom properties, meaning you can alter them inline.
+The `global-layout-with-sidebar` component’s settings are mapped to CSS custom properties, meaning you can alter them inline. You can also set custom properties within your stylesheet and modify them using media queries to adjust for different view widths.
 
 In the following example, the `gap` between the columns is `2em`, the 1-column layout is triggered when the main content goes under `66.666%`, and the sidebar is `300px` wide in the 2-column layout.
 


toolkits/global/packages/global-layout-with-sidebar/README.md
```diff
@@ -51,7 +51,7 @@ Somewhat unconventional, but the component can take it!
 
 ### Custom settings
 
-The `global-layout-with-sidebar` component’s settings are mapped to CSS custom properties, meaning you can alter them inline.
+The `global-layout-with-sidebar` component’s settings are mapped to CSS custom properties, meaning you can alter them inline. You can also set custom properties within your stylesheet and modify them using media queries to adjust for different view widths.
```
tulipdexter, line 54: I find it strange that we're encouraging overriding these values with inline styles 🤔. 
Would it be better not to mention inline at all, e.g. `meaning you can override them in your stylesheets`, and then provide example scss code?

alexkilgour: what is the use case for being able to override these? and agree and example would be useful

sky-jack: I am also not quite sure of the use case for adjusting these styles inline.

sturobson: I think having the ability to change the value for `--with-sidebar--basis` in. your projects stylesheet to override the default value. You can also change `--with-sidebar--gap` if needed too. 

As the component is [an intrinsic layout](https://www.webdesignerdepot.com/2018/05/a-simple-introduction-to-intrinsic-web-design/) (where you [offer the browser guidance rather than strictly dictate](https://buildexcellentwebsit.es) what you want it to do) I don't think adding a media query is in the spirit of what the component is set out to do,

PR #878, opened 2023-05-16
Removes block spacing styling from Frontend Toolkits as a breaking change

benjclark:
We had a discussion in #frontend recently about block-spacing styles:
https://springernature.slack.com/archives/C0GDUEGMS/p1683886608173249

The conclusion was that we do not think the block-spacing styles we have in Frontend Toolkits and the new Elements are working for us. This PR therefore removes them from Frontend Toolkits as a breaking change.

As part of the Darwin work we are doing at the moment we currently have a situation on Oscar where a single set of HTML is used to display a page that uses styles imported from FT (if a feature flag is disabled) and that same HTML is used to display a page that uses styles imported from Elements (if a feature flag is enabled). This means it is important for us that block-spacing styles are removed from both FT and Elements. We have already removed block spacing styles from the copy of Elements we keep in Oscar, therefore removing from FT helps us with our immediate issue.


sturobson (dismissed):

sturobson:
> Had to revert the update to the global details component.

Ah yes, the old nemesis of installing the package to test the package to commit the package.

PR #882, opened 2023-06-29
Updates the breadcrumb visibility on mobile

nickcall:
To bring both the elements and frontend toolkit breadcrumbs in line, I created this PR to hide all but the last two levels of breadcrumb on small screen devices.

### DIFF
--- a/toolkits/global/packages/global-breadcrumbs/HISTORY.md
+++ b/toolkits/global/packages/global-breadcrumbs/HISTORY.md
@@ -1,7 +1,7 @@
 # History
 
-## 5.1.0 (2023-06-29)
-    * Add new styles for small screen devices to only show last two levels
+## 6.0.0 (2023-06-29)
+    * BREAKING: Add new styles for small screen devices to only show last two levels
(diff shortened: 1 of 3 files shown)


tulipdexter:
Is this breaking? I would consider it a feature (minor)

nickcall:
Yeah, I had a similar chat with @benjclark. We decided it was breaking as it affects and changes the user experience. Happy to go with the majority as it kind of sits in between a bit.

benjclark:
Yeah I'm always a little unsure how to apply semver to frontend changes, particularly CSS changes

In my mind if a team implemented a design ok, and then a version change came along that significantly changes (breaks) that visual design then that version should be a "breaking" change

tulipdexter:
Personally I think that because the change doesn't really affect a part of the UI that the user _depends_ upon, then it's not really a major.

nickcall:
OK, looking at other "breaking" changes in this component, it doesn't feel like this fits the criteria for that so I've gone back to minor.

tulipdexter (approved):

PR #885, opened 2023-07-26
Version 1.0.0 of CSAT

benjclark:
The Customer Satisfaction Input (hereafter referred to as the CSAT) component was an outlier because it had a dependency on another component: Global Forms. In Frontend-Toolkits we advise not to implement dependencies between components.

Because of that component to component dependency consumers of CSAT component had to do some bad things such as define static data for Global Forms in their application in order for CSAT component to render. This was not good and in the context of this component being rolled out across many applications the development effort and maintenance effort of doing that is significant.

In addition to that component to component dependencies in Frontend-Toolkits are troublesome in general. Because it means the consumer application needs to implement both exact versions (CSAT component version and Global Forms version) in their application which can be restrictive and risks things easily getting out of sync accidentally.

This PR simplifies the CSAT component by removing the dependency on Global Forms. Instead the form element HTML is hardcoded into this component and the CSAT component provides styling for that form element HTML. This is considered the lesser evil compared to continuing to depend on Global Forms.

The PR makes other changes in addition, to bring it up to date and introduces new functionality, as outlined in the HISTORY file:
```
## 1.0.0 (2023-07-26)
    * BREAKING
        * Brings component up to date with v32.0.0 of Brand Context
            * Note: Please ensure your application also uses v32.0.0 of Brand Context
        * Removes dependency on Global Forms component. 
            * Note: Please uninstall Global Forms if you do not use elsewhere in your application.
            * Note: The data model needed for this component has been greatly simplified. Please consider updating your application view model where needed.
        * Introduces new data field: id. 
            * Note: Please refer to docs for more info. Please consider updating your application view model if needed.
        * Removes CSAT component design tokens. 
            * Note: Please update your import path from "00-tokens/default.tokens.scss" to "10-settings/default.scss"
        * Major refactor of component's HTML and Sass. 
            * Note: If you do not consume the HTML directly from the installed package please update the HTML in your application.
        * New functionality: now appends responseRating URL param to survey question link href. Refer to the docs for more info.
```

**If you currently use the CSAT component in your application please read the HISTORY file and update to v1.0.0 at your earliest convenience.**


Sreilys (approved):
You might want to get a review from some developer that is using this component and has a direct benefit from those changes

PR #886, opened 2023-07-27
[AK] Add spellcheck attribute so it can be used as input parameter

andreaskoscinski:
To prevent word suggestions from the system/keyboard **spellcheck="false"** can be added to the input field. Due to the fact spellcheck isn't listed in the globalFormAttributes it couldn't be added when using global forms. 

image

This shows the issue, when searching for UCL and hitting enter, results for ICL where listed.

To reproduce this, check https://staging-wayf.springernature.com/?redirect_uri=https%3A%2F%2Frd.springer.com%2Farticle%2F10.1007%2Fs10488-015-0664-7 on Safari.

### DIFF
--- a/toolkits/global/packages/global-forms/HISTORY.md
+++ b/toolkits/global/packages/global-forms/HISTORY.md
@@ -1,4 +1,6 @@
 # History
+## 6.0.1 (2023-07-31)
+    * Adds support for spellcheck attribute
 
 
 ## 6.0.0 (2023-01-18)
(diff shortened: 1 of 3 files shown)


andreaskoscinski:
In case there needs to be more code changed (to merge), just let me know.

Sreilys (approved):
Is inside the if, so looks good to me.

toolkits/global/packages/global-forms/HISTORY.md
```diff
@@ -1,4 +1,6 @@
 # History
+## 6.0.1 (2023-07-31)
+    * Adds support for spellcheck attribute
```
andreaskoscinski, line 5: This empty line is removed but not shown in the changed files diff view

Sreilys (approved):

PR #887, opened 2023-08-14
Version 1.0.1 of Global Customer Satisfaction Input component

benjclark:
Removes aria-required and aria-invalid from CSAT html as not needed

The HTML elements in question already have `required` and/or `invalid` attributes on them and are the correct semantic elements


benjclark:
Lighthouse report after change:


image

PR #890, opened 2023-09-18
Fix `global-customer-satisfaction-input` overflow bug

danielclubb:
Issue: springernature/frontend-toolkits/issues/889

- Add `position: relative` rule to `.c-customer-satisfaction-input`

### DIFF
--- a/toolkits/global/packages/global-customer-satisfaction-input/HISTORY.md
+++ b/toolkits/global/packages/global-customer-satisfaction-input/HISTORY.md
@@ -1,5 +1,8 @@
 # History
 
+## 1.0.3 (2023-09-18)
+    * Fixes issue with visually hidden elements causing an overflow due to absolute positioning.
+
 ## 1.0.2 (2023-08-24)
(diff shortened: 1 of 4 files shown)


danielclubb:
Hey @benjclark,

Here's the cleanest solution I could come up with, just adding a `position: relative` to the parent `.c-customer-satisfaction-input` element makes the elements with absolute positioning become relative to the parent 👍 

I've run `npm run demo -- --p global-customer-satisfaction-input` to test the fix, and have run tests + linting, but I'm not sure if there's any other steps I need to take so if you could let me know that'd be great,

Thanks!
Dan

benjclark:
@danielclubb Lovely stuff. A nice simple fix is all that is needed. Thanks Dan.

All you need to do now is update the HISTORY.md file for the component and increment the patch version number in the package.json file.

Once that is committed I will be able to approve and merge this PR

Thanks so much for looking into this!!

danielclubb:
@benjclark all done 😃 let me know if there's anything else you need,

Thanks for your help!

benjclark (approved):

PR #892, opened 2023-11-10
Add user journeys and additional info to survey link

RSeidelsohn:
In OASIS, we use Survey Monkey to send our customers to an additional survey where they can tell us why they chose the rating they selected.
In these surveys, we additionally want to see the `userJourney` and the `additionalInfo` parameters we add to the customer satisfaction Handlebars template.

For that reason, I enhanced the javascript of the global customer satisfaction tool so that it checks for either parameter's existence and if they are not empty, appends them to the searchParams of the generated survey URL.

Additionally, I made the existing Jest test for that component a bit more flexible and added tests for the two new parameters. Also, I added documentation for these additions to the README, where I removed a duplicated section that was there before.


toolkits/global/packages/global-customer-satisfaction-input/HISTORY.md
```diff
@@ -1,5 +1,8 @@
 # History
 
+## 1.3.0 (2024-11-10)
```
benjclark, line 3: I think it makes sense to jump to 1.1.0, not 1.3.0 ?

toolkits/global/packages/global-customer-satisfaction-input/js/customer-satisfaction-input.js
```diff
@@ -81,22 +134,33 @@ class CustomerSatisfactionInput {
 			event.preventDefault();
 			event.stopPropagation();
 		});
-		['click', 'keydown'].forEach(eventType => {
+
+		const eventTypes = ['click', 'keydown'];
+
+		// ['click', 'keydown'].forEach((eventType) => {
```
benjclark, line 140: Can remove this comment?

benjclark (approved):

PR #905, opened 2026-07-27
Simplify rating description, remove overwhelming information.

raquel-roldo:
Hey! Hopefully this is how it's done, let me know what you think.

### DIFF
--- a/toolkits/global/packages/global-customer-satisfaction-input/view/customerSatisfactionInput.hbs
+++ b/toolkits/global/packages/global-customer-satisfaction-input/view/customerSatisfactionInput.hbs
@@ -3,52 +3,47 @@
 		<form class="c-customer-satisfaction-input__form">
 			<h{{headingLevel}} class="c-customer-satisfaction-input__heading">How was your experience today?</h{{headingLevel}}>
 			<fieldset class="c-customer-satisfaction-input__fieldset c-customer-satisfaction-input__field" id="csat-radios{{#if id}}-{{id}}{{/if}}" name="csat-radios" invalid="" aria-describedby="csat-radios-error{{#if id}}-{{id}}{{/if}}" required="" autocomplete="off">
-				<legend class="c-customer-satisfaction-input__visually-hidden">Rating. A scale of 5 feelings conveyed using images that range from awful to great. The feelings represent how you feel about your experience today.</legend>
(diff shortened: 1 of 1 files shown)


toolkits/global/packages/global-customer-satisfaction-input/view/customerSatisfactionInput.hbs
```diff
@@ -3,52 +3,47 @@
 		<form class="c-customer-satisfaction-input__form">
 			<h{{headingLevel}} class="c-customer-satisfaction-input__heading">How was your experience today?</h{{headingLevel}}>
 			<fieldset class="c-customer-satisfaction-input__fieldset c-customer-satisfaction-input__field" id="csat-radios{{#if id}}-{{id}}{{/if}}" name="csat-radios" invalid="" aria-describedby="csat-radios-error{{#if id}}-{{id}}{{/if}}" required="" autocomplete="off">
-				<legend class="c-customer-satisfaction-input__visually-hidden">Rating. A scale of 5 feelings conveyed using images that range from awful to great. The feelings represent how you feel about your experience today.</legend>
+				<legend class="c-customer-satisfaction-input__visually-hidden">Select one rating.</legend>
```
morgaan: I do wonder if we cannot even simplify further down to just “Rating” TBH.

Reasoning:

1. Some screen readers (e.g. Voice over) will read the legend out for every field what may be cumbersome and redundant to hear the "Select one" prefix each time.
2. The user is presented with radio buttons what may implicitly suggests it is a single-choice option, and therefore make the “Select one” redundant.

That being said, I had seen examples by accessibility specialists including the action verb (e.g. select, choose...) in the `legend` 🤷.

It may be interesting to get @fred-warburton's view on this.

morgaan (commented):
Thank you for tackling this @raquel-roldo 🫶

I left a comment for which I'd be interested to get @fred-warburton view on.

toolkits/global/packages/global-customer-satisfaction-input/view/customerSatisfactionInput.hbs
raquel-roldo: Thanks @morgaan. That's a good point, I'd go with just "Rating", specially considering the possibility of it being read multiple times by some screen readers.

